PR #3, opened 2019-11-02
Fix typo in comment for pattern of L in ldl code

mjacobse:
I believe there is a small typo in this comment for the ldl code. From what I understand, `Pattern` holds the column indices of the k-th row of L, not the row indices of the k-th column of L.


DrTimothyAldenDavis:
Yes, that looks like a typo.  It's a very minor change but just to keep things consistent for all contributions to SuiteSparse, can you read over the CONTRIBUTING.md file and follow those instructions first?  Then I can do the merge.

DrTimothyAldenDavis:
Contributor agreement not signed.

PR #4, opened 2019-11-13
Fix compilation on GNU/Hurd

svillemot:
This is a patch applied on the Debian package.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=926347 for more background.

### DIFF
--- a/GraphBLAS/Demo/Include/simple_timer.h
+++ b/GraphBLAS/Demo/Include/simple_timer.h
@@ -36,7 +36,7 @@
 
 #include <time.h>
 
-#if defined ( __linux__ )
+#if defined ( __linux__ ) || defined ( __GNU__ )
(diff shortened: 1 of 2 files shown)


DrTimothyAldenDavis:
Thanks for the suggestion.  Before I do the pull request, can you sign the contributor agreement and email it back to me?

svillemot:
As you can see from the Debian bug log, Paul Sonnenschein is the author of the patch, so I guess you should rather contact him.

DrTimothyAldenDavis:
Thanks for the update!  I've done it in a slightly different way, to make the #ifdefs consistent between simpler_timer.c and simpler_timer.h, in GraphBLAS v3.2.1 (just released in the GraphBLAS git repo).  I'll post here in SuiteSparse shortly.

PR #7, opened 2019-11-19
set install_name on mac

minrk:
install_name is the mac equivalent of soname, and is required for dependants to find suitesparse libs after they are upgraded but still ABI compatible (i.e. SO_VERSION is unchanged). In short, everywhere soname is set on linux, install_name should be set on darwin.

Prevents errors like:

```
Library not loaded: @rpath/libsuitesparseconfig.4.5.3.dylib
  Referenced from: /Users/minrk/conda/envs/petsc/lib/libslepc.3.7.dylib
  Reason: image not found
```

which are caused when a lib is compiled against 4.5.3 and suitesparse is upgraded to e.g. 4.5.4, which is ABI compatible and should work without modification. The name in the "Library not loaded" line is set by install_name.

See https://github.com/conda-forge/suitesparse-feedstock/issues/22 for this issue tracked in the conda-forge community.

### DIFF
--- a/GraphBLAS/Tcov/Makefile
+++ b/GraphBLAS/Tcov/Makefile
@@ -35,7 +35,7 @@ ifeq ($(UNAME),Darwin)
     # Mac
     CFLAGS += -fno-common
     SO_NAME = libgraphblas_tcov.dylib
-    SO_OPTS += -dynamiclib -shared -undefined dynamic_lookup
+    SO_OPTS += -dynamiclib -shared -Wl,-install_name -Wl,$(SO_NAME) -undefined dynamic_lookup
 else
     # Linux
     SO_NAME = libgraphblas_tcov.so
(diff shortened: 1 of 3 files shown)


minrk:
CLA signed and submitted via email

minrk:
conflict resolved

DrTimothyAldenDavis:
Thanks!  Sorry I neglected to include this in 5.7.1.  It will be in the v5.7.2 formal release, and of course on the latest version now.

matthewberryman:
This change breaks clean install on macOS for me with the following error:
```
[100%] Built target mongoose_dylib
dyld: Library not loaded: libsuitesparseconfig.5.dylib
  Referenced from: /Users/matthew/code/SuiteSparse/Mongoose/build/./bin/demo
  Reason: image not found
/bin/sh: line 1: 79626 Abort trap: 6           ./bin/demo
make[1]: *** [default] Error 134
make: *** [go] Error 2
```

PR #8, opened 2019-11-19
allow override of BLAS, LAPACK

minrk:
use `?=` to define defaults instead of hardcoding values

enables selecting blas implementation such as openblas, etc. on mac, among other things

conda-forge [uses this](https://github.com/conda-forge/suitesparse-feedstock/blob/b0538b82b67f226a8aadca07d0665e04c9a7a0d5/recipe/build.sh#L26-L27) to build once against netlib and run against either openblas or mkl.

### DIFF
--- a/SuiteSparse_config/SuiteSparse_config.mk
+++ b/SuiteSparse_config/SuiteSparse_config.mk
@@ -171,12 +171,12 @@ SUITESPARSE_VERSION = 5.6.0
             #   $(MKLROOT)/lib/intel64/libmkl_intel_thread.a \
             #   -Wl,--end-group -lpthread -lm
             # using dynamic linking:
-            BLAS = -lmkl_intel_lp64 -lmkl_core -lmkl_intel_thread -liomp5 -lpthread -lm
-            LAPACK =
+            BLAS ?= -lmkl_intel_lp64 -lmkl_core -lmkl_intel_thread -liomp5 -lpthread -lm
+            LAPACK ?=
         else
             # use the OpenBLAS at http://www.openblas.net (CAN BE VERY SLOW;
             # CHOLMOD can be slowed down by a factor of 100x in extreme cases)
-            BLAS = -lopenblas
+            BLAS ?= -lopenblas
         endif
     endif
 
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Thanks -- very helpful update.

matthewberryman:
This has broken compilation for me:
```
ranlib libcholmod.a
/usr/bin/ld: cannot find -llapack
collect2: error: ld returned 1 exit status
Makefile:541: recipe for target '/home/matthew/code/SuiteSparse/lib/libcholmod.so.3.0.14' failed
make[3]: *** [/home/matthew/code/SuiteSparse/lib/libcholmod.so.3.0.14] Error 1
```

OS: Ubuntu 18.04.4
Intel MKL 2020 Initial Release (installed via
```shell
MKLVERSION=2020.0-088

curl https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2019.PUB | sudo apt-key add -
sudo sh -c 'echo deb https://apt.repos.intel.com/mkl all main > /etc/apt/sources.list.d/intel-mkl.list'
sudo apt-get update && sudo apt-get -y upgrade
# install latest MKL
sudo apt-get -y install intel-mkl-$MKLVERSION
sudo sh -c 'echo /opt/intel/lib/intel64 > /etc/ld.so.conf.d/mkl.conf'
sudo sh -c 'echo /opt/intel/mkl/lib/intel64 >> /etc/ld.so.conf.d/mkl.conf'
sudo ldconfig
```

MKLROOT=/opt/intel/compilers_and_libraries_2020.0.166/linux/mkl
set via
```shell
[ -f /opt/intel/mkl/bin/mklvars.sh ]  && source /opt/intel/mkl/bin/mklvars.sh
```

The problem to me seems to be that this line is sufficient for both BLAS and LAPACK from Intel MKL
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/443558a8a7fd3081d37c2dd1563d9b4536f4b6e4/SuiteSparse_config/SuiteSparse_config.mk#L174
however the combination of both https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/443558a8a7fd3081d37c2dd1563d9b4536f4b6e4/SuiteSparse_config/SuiteSparse_config.mk#L162 combined with the modified line https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/443558a8a7fd3081d37c2dd1563d9b4536f4b6e4/SuiteSparse_config/SuiteSparse_config.mk#L175 results in LAPACK=-llapack which isn't required for MKL and thus generates the error.

I suggest that https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/443558a8a7fd3081d37c2dd1563d9b4536f4b6e4/SuiteSparse_config/SuiteSparse_config.mk#L175 be changed to read `LAPACK=`
with perhaps a comment that the BLAS variable alone is sufficient to pick up LAPACK from MKL.

PR #11, opened 2019-11-23
Fixed typo in Makefile

schmidtp1:
Fixed typo in Makefile that causes installation error


DrTimothyAldenDavis:
Thanks for the suggestion.  I fixed it another way with v5.7.1.

PR #18, opened 2020-01-28
 Fix bad check on allocation return value

ushachar:
Mainly to clean up the compiler warning on missing parans.


DrTimothyAldenDavis:
Thanks -- I saw that.  I've already fixed that in my internal draft (not
yet posted here on github) so it will be applied once I post GraphBLAS
3.2.0.

DrTimothyAldenDavis:
This is now fixed with GraphBLAS v3.2.0.

PR #22, opened 2020-02-29
set linker flags for LAPACK a bit differently

matthewberryman:
addresses problem identified in https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/8#issuecomment-591779080 in a slightly better manner than I proposed there, IMHO


DrTimothyAldenDavis:
Thanks for the update!

PR #24, opened 2020-03-06
UMFPACK: update version in header to latest (5.7.9)

alexeicolin:
Fixes reported version mismatch since 6ee9c7e3d3a6e8fa1322abce8cafd0dc0399d66a
(SuiteSparse v5.5.0) when the UMFPACK source was updated to 5.7.9 but version
in the header stayed at 5.7.8.

This version information is reported by UMFPACK at runtime, e.g.:

    UMFPACK V5.7.8 (Nov 9, 2018), Info:

And mismatch between the reported versions and the version in the source tree
(in ChangeLog) leads to confusion for users with regards to which UMFPACK
library is being used in build of software that links against UMFPACK.

I thought it's worth submitting this PR because SuiteSparse is released more frequently than UMFPACK, so the mismatch can be fixed in next SuiteSparse release.

To avoid the contributor agreement process, instead of merging this PR maybe just make the edit on your end.

### DIFF
--- a/UMFPACK/Include/umfpack.h
+++ b/UMFPACK/Include/umfpack.h
@@ -95,7 +95,7 @@ extern "C" {
 /* Version, copyright, and license */
 /* -------------------------------------------------------------------------- */
 
-#define UMFPACK_VERSION "UMFPACK V5.7.8 (Nov 9, 2018)"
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Thanks for pointing it out.  I'll go ahead and fix it now.

PR #26, opened 2020-03-15
UMFPACK: avoid warning with SunStudio compiler

jedbrown:
```
"../Source/umfpack_qsymbolic.c", line 2693: warning: argument #7 is incompatible with prototype:
        prototype: pointer to function(int, int, int, pointer to int, pointer to int, pointer to int, pointer to void, pointer to double) returning int : "../Source/umfpack_qsymbolic.c", line
 660
        argument : pointer to void
```

This is a very simple change. Let me know if you need a CLA.

### DIFF
--- a/UMFPACK/Source/umfpack_qsymbolic.c
+++ b/UMFPACK/Source/umfpack_qsymbolic.c
@@ -2690,8 +2690,8 @@ GLOBAL Int UMFPACK_qsymbolic
         Quser,
 
         /* no user provided ordering function */
-        (void *) NULL,
-        (void *) NULL,
+        NULL,
+        NULL,
 
         SymbolicHandle, Control, User_Info)) ;
 }


paulmenzel:
It looks like you [have to print a text file, sign it and email it](https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/master/CONTRIBUTING.md).

DrTimothyAldenDavis:
I'm constrained by my organizational policy to require a signed contributor agreement before I can merge any PR, even minor ones.

jedbrown:
An issue with CLAs is that people with blanket authorization to affirm the [DCO](https://developercertificate.org/) on open source contributions usually have to route CLAs through institutional legal offices, especially if those CLAs use any [non-standard wording](https://opensource.com/article/19/1/cla-proliferation). I encourage reading [Richard Fontana's article on CLAs](https://opensource.com/article/19/2/cla-problems). Note that [inbound=outbound is explicit in GitHub's TOS](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#6-contributions-under-repository-license).

DrTimothyAldenDavis:
I understand, but I can't change the policy on my side, and inbound=outbound won't work.  Is there a set of canonical CLAs?  If mine has non-standard wording, perhaps I could ask to use a standard CLA with the same intent as the one I'm currently constrained to use.

jedbrown:
You could accept contributions inbound as MIT/Apache-2.0, which is compatible with all your outbound licenses. Yes, in theory, someone could extract specific code lines (that had been merged under these terms) for use in a more permissive distribution, but I think it would be easier for someone with the skills to assemble a "competitor" (filling in the >90% that is not licensed in that way) to write the whole thing from scratch (as Jack Poulson did with Catamari). I understand if you can't/don't want to consider this.

PR #46, opened 2020-06-17
Mark 'cs' as potentially unused variable

varunagrawal:
Newer versions of gcc throw a warning that the `cs` variable is unused.

```
ccolamd.c:1563:9: warning: variable 'cs' set but not used [-Wunused-but-set-variable]
     Int cs ;
         ^~
```

This PR marks the variable as potentially unused so that the compiler no longer complains about it.

### DIFF
--- a/CCOLAMD/Source/ccolamd.c
+++ b/CCOLAMD/Source/ccolamd.c
@@ -1560,7 +1560,7 @@ PUBLIC Int CCOLAMD_2	    /* returns TRUE if successful, FALSE otherwise */
     Int *dead_cols ;
     Int set1 ;
     Int set2 ;
-    Int cs ;
+    Int cs __attribute__((unused)) ;
 
     int ok ;
 


varunagrawal:
@DrTimothyAldenDavis can you please take a look at this. I have already sent in the CLA form via email.

DrTimothyAldenDavis:
The variable is used when debugging is enabled, so flagging it as unused would break the code in that case.

varunagrawal:
So the attribute attached won't break the code, it simply tells the compiler there is a *possibility* that the variable is unused, so that the compiler won't complain.

This is simply a preprocessor directive and is supported by both GCC and Clang.

From the GCC manual:

> unused
> This attribute, attached to a variable, means that the variable is meant to be possibly unused. GCC will not produce a warning for this variable.

DrTimothyAldenDavis:
The code has to be compilable by gcc, icc, clang, MS Visual Studio, etc.  I can't add directives like this.

PR #59, opened 2020-09-12
SLIP_LU: allow custom paths for gmp and mpfr

fabiencastan:
If gmp and mpfr are not installed on the OS, there is no solution to get them used by SLIP_LU.
This PR allows to specify paths to mpfr and gmp with the variables MPFR_ROOT and GMP_ROOT.


DrTimothyAldenDavis:
This is no longer needed since I'm now switching to a cmake-based build system.

PR #61, opened 2020-09-15
Update ldl.h

FabienPean:
* Add header guards
* C++ friendliness


DrTimothyAldenDavis:
Thanks for the suggestion -- it's something I should add.  But I'd need a signed contributor agreement first before I can merge the PR.  I realize it's a minor change, but I'm constrained by the policy by my organization.

I've made other changes to LDL since then so I'll go ahead a make the related changes myself.

PR #67, opened 2020-11-16
Update CMakeLists.txt

ShuangLiu1992:
fix linking issues on platforms that do not have rt
for example Android is UNIX and NOT Apple but doesn't have rt either.
would be better to have CMake to check whether rt should be linked


paulmenzel (approved):

DrTimothyAldenDavis:
Thanks for the suggestion.  I've removed the need for the rt lib in Mongoose, so this is no longer an issue.  I have a function in SuiteSparse_config, called SuiteSparse_time(), that should be used instead.

PR #91, opened 2021-05-26
Avoid C warnings

gruenich:
Please double check that the generated files are working, as I was not able to regenerate all files with Octave (missing function contains).

### DIFF
--- a/GraphBLAS/Source/GB_math.h
+++ b/GraphBLAS/Source/GB_math.h
@@ -564,6 +564,7 @@ inline float GB_frexpef (float x)
 {
     int exp ;
     float mantissa_ignored = frexpf (x, &exp) ;
+    (void) mantissa_ignored ; // to silence unused variable warnings
     return ((float) exp) ;
 }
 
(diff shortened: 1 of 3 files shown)


DrTimothyAldenDavis:
Thanks for the update.  Those can work, but I think it would be better to just delete those unused variables, and to delete the commented-out lines in the *defs.h file.   I've actually  already deleted the comments in the *defs.h file in my master branch on the GraphBLAS github, but the code isn't released as a tagged release yet on the stable GraphBLAS branch. 

For GraphBLAS, I do all my work on the GraphBLAS github repo, not the overall SuiteSparse one, and then I copy over each stable release to this github repo when I'm ready to integrate the latest GraphBLAS into SuiteSparse.  So for GraphBLAS updates, it's best to do a PR on the GraphBLAS github instead of this one (although I can also take a PR on this github and then copy over the changes to the GraphBLAS github as well).

The GraphBLAS stable release is now at v5.0.6, while SuiteSparse/GraphBLAS is v5.0.5.

 I'll go ahead and merge this pull request, since the changes are easy to incorporate/modify into my development ('master') branch, and then revise accordingly.

DrTimothyAldenDavis:
I also plan to revise my MATLAB scripts for generating the source, and for the test suite, so I can use Octave and MATLAB equally.  The "contains" function in the codegen_*m scripts can be replaced with strfind, which Octave has.

PR #111, opened 2021-11-16
Apply Windows patches we use to build SuiteSparse for Julia on Windows

### LINKED ISSUE #90: Incorporating a patch for building shared libraries that we carry in Julia
We would love to incorporate these two patches into SuiteSparse if possible, which we apply when we build it for Julia:

As discussed in https://github.com/JuliaLang/julia/issues/37323, this patch that makes it possible for us to build shared libraries:
https://github.com/JuliaLang/julia/blob/master/deps/patches/SuiteSparse-shlib.patch

If these can be accepted, I am happy to make a PR.

[VS EDIT: I had originally linked 2 patches - but I removed one of them which is no longer required]

ViralBShah:
As discussed in #90 

Fix #90

### DIFF
--- a/SuiteSparse_config/SuiteSparse_config.mk
+++ b/SuiteSparse_config/SuiteSparse_config.mk
@@ -357,8 +357,8 @@ SUITESPARSE_VERSION = 5.10.1
     #---------------------------------------------------------------------------
 
     ifeq ($(UNAME),Linux)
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis (approved):

DrTimothyAldenDavis:
Got it -- thanks.  I'll test it out and make sure it works for me, and then I'll include this in my next official stable release.

PR #114, opened 2022-01-04
Fix Linking error when run executables in SLIP_LU/Demo

ByLiZhao:
The linker can not find dynamic libraries in ../../lib w.r.t. the
'SLIP_LU/Demo' folder. By passing the '-L' option to GCC, which GCC
ultimately passes to the linker, all executables are linked correctly.
But when these executables invoked from Make, they have trouble find
dynamic libraries in the said folder, because the folder is not in the
list of search paths of the linker.

To solve this problem, an extra '-Wl, rpath=../../lib' option is passed
to GCC. This tells the linker to set the DT_RPATH attribute in the
.dynamic section of the resulted executables. When the executables are
invoked, they will first try to find dynamic libraries in the path
indicated by DT_RPATH before looking some standard library paths.


DrTimothyAldenDavis:
I'm now switching to a cmake-based build system so this Makefile is gone.

PR #123, opened 2022-03-23
Install mongoose dll

xantares:
This is already applied upstream


DrTimothyAldenDavis:
Thanks for the PR.  This should be fixed in the new cmake build system for all of SuiteSparse.  Can you give it a try and let me know if it does what you expect?  I'm not able to test any of my packages on Windows.

xantares:
fixed, thanks

PR #131, opened 2022-06-27
SPQR documentation: compatibility fix for TeX Live 2022

svillemot:
The \htmladdnormallink macro has been removed from hyperref as of TeX Live
2022. It was simply an alias for \href (with arguments swapped), so use the
latter instead.


DrTimothyAldenDavis:
THanks for the suggestion.  I've switched to using  the \url command instead.

Issue #138, opened 2022-10-16
Incorrect pointer assignment in umf_kernel_init 


advanpix:
Pointers Rpi & Rpx are assigned to incorrect memory blocks.

```
rpi = UMF_mem_alloc_tail_block (Numeric, UNITS (Int *, n_row+1)) ;
rpx = UMF_mem_alloc_tail_block (Numeric, UNITS (Entry *, n_row+1)) ;
...
Rpi = (Int   **) (Memory + rpx) ; <--
Rpx = (Entry **) (Memory + rpi) ; <--
```

Should be:
```
Rpi = (Int   **) (Memory + rpi) ;
Rpx = (Entry **) (Memory + rpx) ;
```

The bug reveals itself if `UNITS (Int *, n_row+1) != UNITS (Entry *, n_row+1)`.

DrTimothyAldenDavis:
This proposed fix does correct the confusion between Rpi and Rpx, but https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/2a35c27b9f2d50383501e7f62027b9098707ac21 is the complete fix.

DrTimothyAldenDavis:
Thanks for the feedback!

advanpix:
> This proposed fix does correct the confusion between Rpi and Rpx, but [2a35c27](https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/2a35c27b9f2d50383501e7f62027b9098707ac21) is the complete fix.

Well, this is interesting. You just took the results of my work and made commit under your name. 
Basically you just stripped my authorship to make my contribution to the project untraceable.
This is the worst policy of accepting the contributions in open-source world I have ever seen.

You should clearly state this on the main page of the repository.

DrTimothyAldenDavis:
Your proposed fix was in the right direction, but broken, so I couldn't merge it as-is so I fixed the problem in a different way.  I'm happy to credit you for pointing out the problem.  I do have a contributor policy posted (which wasn't followed).  I require a signed contributor agreement for all PRs.  That requirement is stated when you do a pull request.

I'll back out the changes.  If you'd like to propose them, please sign a contributor agreement and do another PR.

PR #139, opened 2022-10-16
Fixes DrTimothyAldenDavis/SuiteSparse#138

advanpix:
Fix for the #138 issue and small performance improvement

### DIFF
--- a/UMFPACK/Source/umf_kernel_init.c
+++ b/UMFPACK/Source/umf_kernel_init.c
@@ -522,8 +522,8 @@ GLOBAL Int UMF_kernel_init
 	/* out of memory, which can only mean that the pattern has changed */
 	return (FALSE) ;	/* pattern changed */
     }
-    Rpi = (Int   **) (Memory + rpx) ;
-    Rpx = (Entry **) (Memory + rpi) ;
+    Rpi = (Int   **) (Memory + rpi) ;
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
The test "n1 > 0 && drop" is incorrect.  The pruning must always occur since it always prunes exact zeros from the singleton rows/columns.  "drop" is true if in addition to exact zeros, small entries are to be checked for and dropped if present.  But I always remove exact zeros and that shouldn't be changed.

I think the pointer issue is a good addition.  It's not actually a problem, because both memory spaces must be the same size in all cases (both hold pointers, one pointing to an Int and the other to an Entry).  That is, even if sizeof(Int) and sizeof(Entry) differ (they often do), the sizeof(Int *) and sizeof(Entry *) should always be identical.

But the code looks confusing and should be revised, as you point out in #138.  However, to add that change I would need a signed contributors agreement first (a signed PDF is fine, with a digital signature).  Would that be OK?

DrTimothyAldenDavis:
The suggestion to fix the code confusion is a good one, but the fix needs to be a bit different.  The change to rpi and rpx that you suggest also changes the memory layout of the factorization, which I'd rather not change.  I'd rather fix it so that the confusion between (Int *) and (Entry *) is resolved, but in a way that leaves the underyling memory layout unchanged.

PR #167, opened 2022-11-07
Support building with CMAKE_RELEASE_POSTFIX

opoplawski:
Fedora builds 3 versions of suitesparse - 32-bit BLAS and 64-bit BLAS with and without trailing "_"'s on symbols.  I'd like to use CMAKE_RELEASE_POSTFIX to do this, but we need to change the name of the libraries that suitesparse looks for accordingly.

### DIFF
--- a/AMD/cmake_modules/FindAMD.cmake
+++ b/AMD/cmake_modules/FindAMD.cmake
@@ -41,7 +41,7 @@ find_path ( AMD_INCLUDE_DIR
 
 # compiled libraries AMD
 find_library ( AMD_LIBRARY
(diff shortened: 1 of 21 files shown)


DrTimothyAldenDavis:
Looks great -- can you email me a signed contrib agreement?  A electronically-signed PDF is fine (no need to print on paper & scan it in).

DrTimothyAldenDavis:
I'm constrained by my organizational policy to require a signed contributor agreement before I can merge any PR, even minor ones.  I'm not able to change that policy, unfortunately.

PR #169, opened 2022-11-07
make the amd,camd and klu doc makefiles portable between bash and dash

kiwifb:
* Please describe your pull request:
I have been sitting on this since April 2020 https://github.com/cschwan/sage-on-gentoo/issues/574
The gist of it is that the makefile to produce the documentation for amd, camd and klu will fail if you use bash instead of dash. This is because the `echo` is not portable and its behaviour will vary depending on the shell used. See the section on echo at https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/Limitations-of-Builtins.html#Limitations-of-Builtins

This branch fixes this by using `printf` instead of `echo` which will work across bash and dash at least.

### DIFF
--- a/AMD/Doc/Makefile
+++ b/AMD/Doc/Makefile
@@ -11,9 +11,9 @@ default:  AMD_UserGuide.pdf
 
 AMD_UserGuide.pdf: AMD_UserGuide.tex AMD_UserGuide.bib ../Include/amd.h \
     amd_version.tex
-	echo '\\begin{verbatim}' > amd_h.tex
+	printf '\\begin{verbatim}\n' > amd_h.tex
 	expand -8 ../Include/amd.h >> amd_h.tex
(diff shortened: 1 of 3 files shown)


DrTimothyAldenDavis:
Looks great - thanks!

DrTimothyAldenDavis:
Oops.  I should have mentioned:  I do PR's into the dev or dev2 branches (in this case dev2), not master.  No worries:  I'll copy these updates into the dev2 branch (which is a bit ahead of master at the moment).  In the future though it would be easier to do PRs into the dev2 or dev branches.   I do my work there and then when it's ready I do a single PR over to the (default) master branch.

kiwifb:
I would suggest you add a comment about that in the `CONTRIBUTING.md` file.

DrTimothyAldenDavis:
Shall do.

PR #181, opened 2022-11-19
Fix building SPQR with NPARTITION=ON

MehdiChinoune:
Proposed changes

### DIFF
--- a/SPQR/CMakeLists.txt
+++ b/SPQR/CMakeLists.txt
@@ -201,8 +201,14 @@ include_directories ( ${BLAS_INCLUDE_DIRS} )
 
 # cholmod:
 target_link_libraries ( spqr PUBLIC ${CHOLMOD_LIBRARIES} )
+if ( NPARTITION )
+target_compile_definitions ( spqr PRIVATE NPARTITION )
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Thanks for the update.  I'm in the progress of a better fix, where CHOLMOD can be compiled with -DNPARTITION or without, and this has no effect on how SPQR is compiled.  If SPQR requests CHOLMOD for an ordering via METIS, a run-time error will be returned (CHOLMOD_NOT_INSTALLED).  This is easier than requiring SPQR to know how CHOLMOD was compiled.

The -DNPARTITION option will be removed from SPQR, as a result, so I can close this pull request eventually.  I'll leave it open until I post the updated SPQR on the dev2 branch.

DrTimothyAldenDavis:
See the latest version of the dev2 branch, and the most recent push: https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/39ce416ec6ce485862ad5ba19018ee9f235899f2 .  I'm still testing it but it works fine on my Mac, when CHOLMOD is compiled with -DNPARTITION.

PR #184, opened 2022-11-28
Add GitHub action as CI for MinGW

mmuetzel:
This PR adds rules for GitHub hosted runners to build SuiteSparse in different build environments for MinGW using MSYS2.

It is currently configured to run on each push. Other triggers are also possible (e.g., on each pull request).

I hope that will help to keep compatibility with that platform. Thank you very much for your recent efforts to fix some build issues.

One remaining issue is that the libraries don't have version suffices on Windows. So, their names can't be used to determine the version in the .cmake files. Instead, this PR includes changes to parse the corresponding headers for the version. That is strongly "inspired" by the patch in this PR https://github.com/msys2/MINGW-packages/pull/14156.

The ccache size probably still needs adjusting. A cache size of approximately 2x the cache after the first run is a good minimum afaict.

Edit: The GCC environments (MINGW64 and MINGW32) are currently failing to build SPEX because of a downstream issue with mpfr.

Edit 2: See, e.g., here for what that could look like:
https://github.com/mmuetzel/SuiteSparse/actions/runs/3568435121


DrTimothyAldenDavis:
Yes, I think you're right, that running on MINGW would be a good first platform.  I test on Linux and Mac extensively.  I rarely test on Windows, if at all.

I will often be working on many different machines at the same time (Linux on a laptop, Linux on a desktop, Mac M1, Mac Intel) and I use git to sync between the four systems, even with tiny changes.  Sometimes I push a broken code on purpose.

Can the github action be limited to specific branches?  Say "stable" and "dev"?  Then I can use dev2 for raw changes that are not yet ready for the CI tests.

mmuetzel:
From the point of view of a contributor, it might be good if the CI also ran on custom branches.
What I generally do when I try to contribute on GitHub is:
1. Fork the repository.
2. Create a feature branch.
3. Push the changes to that branch. (Repeat this step as often as needed.)
4. Create a PR.

Imho, it is pretty handy if there is some kind of CI that runs before creating the PR. That way, a contributor could check whether their changes might cause regressions *before* they create a PR. (That might also help you reviewing the changes because you'd at least know that they have a lower risk of breaking compilation.)

But I understand that you'd like to have a branch for which the CI doesn't run.
There is the option to exclude branches from the action with `branches-ignore`:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore

Would it be ok to ignore just `dev2` for that purpose?

DrTimothyAldenDavis (approved):

PR #185, opened 2022-12-03
Add action rules for GitHub-hosted Ubuntu runners.

mmuetzel:
This PR adds rules to build SuiteSparse on GitHub-hosted Ubuntu runners. 
It tries to build with GCC and Clang compilers.

Currently, it fails because there seems to be some mix-up with shared vs. static libraries. See, e.g.:
https://github.com/mmuetzel/SuiteSparse/actions/runs/3608879679/jobs/6081737279#step:7:859
```
-- Found AMD: /home/runner/work/SuiteSparse/SuiteSparse/AMD/build/libamd.a (found suitable version "3.0.2", minimum required is "3.0.2") 
-- AMD version:      3.0.2
-- AMD include dir:  /home/runner/work/SuiteSparse/SuiteSparse/AMD/Include
-- AMD dynamic:      /home/runner/work/SuiteSparse/SuiteSparse/AMD/build/libamd.a
-- AMD static:       /home/runner/work/SuiteSparse/SuiteSparse/AMD/build/libamd.a
```

Are you sure you prefer to not have the CI run on your development branch? Imho, it is totally fine if the CI fails for some period on a development branch. But it might be useful to see the current state, nevertheless.
I don't know your development cycle. But the CI might be useful to check what might need to be fixed before merging your development branch to the stable branch.

I also adjusted the size of the ccache for MinGW in the same commit. I hope that is fine.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,6 +8,88 @@ concurrency: ci-${{ github.ref }}
 
 jobs:
 
+  ubuntu:
+    # For available GitHub-hosted runners, see:
+    # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
+    runs-on: ubuntu-latest
+
+    strategy:
+      # Allow other runners in the matrix to continue if some fail
+      fail-fast: false
+
+      matrix:
+        compiler: [gcc, clang]
+        include:
+          - compiler: gcc
+            compiler-pkgs: "g++ gcc"
+            cc: "gcc"
+            cxx: "g++"
+          - compiler: clang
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
I'm in the middle of revising the static vs dynamic build, on the dev2 branch.  I needed to check some partial work on another machine, so I pushed the current state (in progress, not done).  So the dev2 branch should never be used with CI.

I  plan on setting up a dev branch, and the CI will run on that.  dev2 will be the draft development branch.

So feel free to add to the yaml and such.  I'm still working on the static/dynamic builds in the CMakeLists.txt.  Those will currently break, since it's still in progress.

PR #198, opened 2022-12-10
CI: Skip building SPEX on MINGW64 and MINGW32.

mmuetzel:
This addresses #181. The commit cb9cae8 (CI: Skip building SPEX on MINGW64 and MINGW32.) can probably be reverted when the issue with mpfr has been fixed in MSYS2.
I manually edited the `Makefile` and commented out all lines related to SPEX with a `sed` replacement rule. Is there a more "default" way of selecting which parts of SuiteSparse should be built (or skipped)?

Additionally, this PR increases the ccache size on Ubuntu (which didn't build at the time the CI rule was originally written). I underestimated the size that we were going to need.

In a last change, I removed `-DGBNCPUFEAT=ON` from the build flags on MinGW. IIUC, this is no longer needed now that MinGW is correctly detected by GraphBLAS.

PS: Is the `dev` branch the correct target for PRs now?

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -197,8 +197,7 @@ jobs:
                               -DCMAKE_CXX_COMPILER_LAUNCHER='ccache' \
                               -DCMAKE_Fortran_COMPILER_LAUNCHER='ccache' \
                               -DENABLE_CUDA=OFF \
-                              -DBLA_VENDOR=OpenBLAS \
-                              -DGBNCPUFEAT=ON"
+                              -DBLA_VENDOR=OpenBLAS"
 
       - name: ccache status
         continue-on-error: true
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -180,6 +180,15 @@ jobs:
           ccache -p
           ccache -s
 
+      - name: disable SPEX
+        # SPEX currently doesn't link correctly with the mpfr library in MINGW64 and MINGW32.
+        # So, disable building it for the time being.
+        # Remove this step once the issue is fixed upstream.
+        # See also: https://github.com/msys2/MINGW-packages/pull/14146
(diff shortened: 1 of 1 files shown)


mmuetzel:
Related to this: Are there some checks that could be run in the CI to see whether the compiled binaries actually work as expected?

DrTimothyAldenDavis:
dev2 would be better.   That's always the latest bleeding edge version.  But dev is fine since the 2 branches are identical right now.

I have a range of tests and demos.  The simplest one is a top-level 'make demos' which then does a 'make demos' in each of the packages.  

That target does a "cd build ; cmake -DDEMO=1 .. ; make" to enable the cmake to build the demo programs.  Then the script to run the demos is in the Makefile.  See AMD/Makefile for example.  Building the demos would be easy on Windows, since that is all cmake based (just enable the DEMO flag).  The downside is that scripts to run the demos are in the top-level Makefiles for each package.

Many (not all) of the demos compare their results with known results, say in AMD/Demo/*.out.  However, these demos are not strictly tests.  They don't report failures if the diff shows a difference, however, because the outputs include things like residual norms (which are OK if they are slightly different), and run times (which are always different), and in some cases, the demo outputs include a report of the name of the BLAS/LAPACK library used.

However, I could revise the demos slightly to report a test failure if (say) the residual norm is higher than expected.  I would have to add the expected resid norm for each input matrix, since these demos take in arbitrary matrices and some of those can be arbitrarily ill-conditioned.

The brutal tests are in the SuiteSparse/Package/Tcov folders, but those all require Linux and all use plain Makefiles.  They don't work on my 2 Macs, and they wouldn't work on Windows.  The tests in the Tcov folders are very long and not suitable for a quick "smoke test" to see if the library was compiled OK.  One of the Tcov tests (GraphBLAS) requires MATLAB.

PR #199, opened 2022-12-10
Support cmake generators other than Makefiles.

mmuetzel:
`cmake` supports building with different generators. Most POSIX platforms default to `* Makefiles` generators. However, the native Windows environments of MSYS2 default to `ninja` (which tends to run faster on that platform). Also, a user on any platform can select any of the available generators.

Currently, the build rules only work if the selected generator produces Makefiles (e.g., `"Unix Makefiles"` or `"MSYS Makefiles"`).

This PR proposes changes that would allow to build SuiteSparse with `cmake` generators that are different from `* Makefiles`.
It also switches the CI build rule for MingW to use its default generator (i.e., `ninja`).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -182,8 +182,7 @@ jobs:
 
       - name: build
         run: |
-          make CMAKE_OPTIONS="-G\"MSYS Makefiles\" \
-                              -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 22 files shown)


DrTimothyAldenDavis:
Thanks for the update.  I'll merge it once the current tests finish and include it in v6.0.2 (which I hope to release soon as an official stable release).  A better PR would be to merge this update into the dev2 branch first, so I can give it a try.  I'll go ahead and merge with stable since the 3 main branches (stable, dev, dev2) are currently identical.

PR #220, opened 2022-12-14
Run demos as part of CI

mmuetzel:
As described in #198, it might be useful to run the demos as part of the CI.
Currently, there is no automatic check whether the results are correct (within some tolerance) or even if they execute without crashing. But having the option to check the logs manually is still better than nothing imho.

For the Windows targets, it is necessary to install SuiteSparse before running the demos, or the demo executables would fail to find the necessary libraries. 
An alternative might have been to add the paths to all previously built libraries to the PATH variable. But that would probably be more fragile with respect to any potential changes to the build system.

Since the demos require a Fortran compiler to be built, they cannot be run for the MSYS2/CLANG32 environment.

I also took the opportunity to make the cmake configuration flags on Linux and MinGW more similar to each other.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -84,6 +84,10 @@ jobs:
                               -DCMAKE_Fortran_COMPILER_LAUNCHER='ccache' \
                               -DBLA_VENDOR=OpenBLAS"
 
+      - name: check
+        run: |
+          make demos
+
       - name: ccache status
         continue-on-error: true
         run: ccache -s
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
The few demos that use a Fortran compiler are disabled if there's no Fortran compiler available.  Are you sure you need to disable all the demos on MSYS2/CLANG32?

DrTimothyAldenDavis:
I just merged these to dev2 and then dev.  The CI output looks fine, except for a strange output from COLAMD and CCOLAMD for MINGW32:

```
diff ./Demo/colamd_l_example.out ./build/colamd_l_example.out
2c2
< Column 0, with 3 entries:
---
> Column 0, with 0 entries:
...
```

The problem is probably just the wrong "%ld" format in colamd_l_example.c.  I can fix that.  The rest of the output must be fine since there's no difference in the list of entries that get printed in each column.

The umfpack outputs look ok too but the diff is huge, even where there's no difference.  Maybe it's a DOS txt file vs Unix txt file issue.  This comes up for MING32 and MING64 but not the other cases.

I'll dig through all the logs to see if anything else looks odd.

DrTimothyAldenDavis:
The 32-bit SPQR demos look a bit bizarre (MINGW32).  Some residuals are completely wrong.  Perhaps related to the change to how SPQR defines its integer size (now int64_t).

PR #228, opened 2022-12-24
Fix dllimport/dllexport for libraries with MSVC

mmuetzel:
The `dllimport`/`dllexport` mechanism that is currently used doesn't look correct. The same macros are used for a number of different libraries. But it needs to be dealt with on a per library basis.
E.g., for the functions exported by the `suitesparse_config` library: When a library that links to `suitesparse_config` is compiled, it must `dllexport` its own functions, but `dllimport` the functions from `suitesparse_config`.
Additionally, for static libraries, the `dllimport`/`dllexport` mechanism mustn't be used. (Neither when compiling nor when consuming the static library.)

The proposed changes use dedicated "API" flags for each library and deal with static libraries.

I also took the opportunity to remove the (unused) `SUITESPARSE_COMPILER_MINGW` because MinGW isn't a compiler.
Additionally, I stopped linking to `libm` on Windows in the `CMakeLists.txt` files that are touched by this PR.

I tested with `gcc`/`gfortran` and also made sure that it compiles with `clang`/`flang`. I don't have MSVC. But I hope it should be working with that family of compilers now, too. (At least, for the libraries that were already using those "API" flags.)

MinGW doesn't *need* `dllimport`/`dllexport`. So, I reverted to using that mechanism only for MSVC in a final commit.

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -84,6 +84,8 @@ set_target_properties ( amd PROPERTIES
     SOVERSION ${AMD_VERSION_MAJOR}
     PUBLIC_HEADER "Include/amd.h" )
 
+target_compile_definitions ( amd PRIVATE AMD_LIBRARY )
+
 #-------------------------------------------------------------------------------
 # static amd library properties
 #-------------------------------------------------------------------------------
@@ -95,6 +97,8 @@ if ( NOT NSTATIC )
         C_STANDARD_REQUIRED 11
         OUTPUT_NAME amd
         SOVERSION ${AMD_VERSION_MAJOR} )
+
+    target_compile_definitions ( amd_static PUBLIC AMD_STATIC )
 endif ( )
 
 #-------------------------------------------------------------------------------
@@ -108,7 +112,7 @@ if ( NOT NSTATIC )
 endif ( )
 
 # libm:
-if ( NOT MSVC )
+if ( NOT WIN32 )
     target_link_libraries ( amd PUBLIC m )
     if ( NOT NSTATIC )
         target_link_libraries ( amd_static PUBLIC m )
--- a/AMD/Config/amd.h.in
+++ b/AMD/Config/amd.h.in
@@ -43,10 +43,10 @@ extern "C" {
 #include "SuiteSparse_config.h"
 
 //------------------------------------------------------------------------------
-// importing/exporting symbols on Windows
+// importing/exporting symbols for Microsoft Visual Studio
 //------------------------------------------------------------------------------
 
-#if defined ( _WIN32 )
+#if SUITESPARSE_COMPILER_MSC
 
     // dllimport/dllexport on Windows
     #if defined ( AMD_LIBRARY )
@@ -62,7 +62,7 @@ extern "C" {
 
 #else
 
-    // for other platforms
+    // for other compilers
     #define AMD_PUBLIC extern
 
 #endif
--- a/AMD/Config/amd.h.in
+++ b/AMD/Config/amd.h.in
@@ -42,7 +42,32 @@ extern "C" {
 
 #include "SuiteSparse_config.h"
 
-SUITESPARSE_PUBLIC int amd_order  /* returns AMD_OK, AMD_OK_BUT_JUMBLED,
+//------------------------------------------------------------------------------
+// importing/exporting symbols on Windows
+//------------------------------------------------------------------------------
+
+#if defined ( _WIN32 )
+
+    // dllimport/dllexport on Windows
+    #if defined ( AMD_LIBRARY )
+        // compiling SuiteSparse itself, exporting symbols to user apps
+        #define AMD_PUBLIC extern __declspec ( dllexport )
+    #elif defined ( AMD_STATIC )
+        // compiling static library, no dllimport or dllexport
+        #define AMD_PUBLIC extern
+    #else
+        // compiling the user application, importing symbols from SuiteSparse
+        #define AMD_PUBLIC extern __declspec ( dllimport )
+    #endif
+
+#else
+
+    // for other platforms
+    #define AMD_PUBLIC extern
+
+#endif
+
+AMD_PUBLIC int amd_order  /* returns AMD_OK, AMD_OK_BUT_JUMBLED,
                                     * AMD_INVALID, or AMD_OUT_OF_MEMORY */
 (
     int32_t n,                     /* A is n-by-n.  n must be >= 0. */
@@ -53,7 +78,7 @@ SUITESPARSE_PUBLIC int amd_order  /* returns AMD_OK, AMD_OK_BUT_JUMBLED,
     double Info [ ]         /* output Info statistics, of size AMD_INFO */
 ) ;
 
-SUITESPARSE_PUBLIC int amd_l_order  /* see above for description */
+AMD_PUBLIC int amd_l_order  /* see above for description */
 (
     int64_t n,
     const int64_t Ap [ ],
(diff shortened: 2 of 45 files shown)


DrTimothyAldenDavis:
Ouch ...  Windows makes this very difficult indeed.

A few questions:  the only library functions that I have in Fortran are in AMD (amd.f and amdbar.f).  I'm concerned that the #if ... #endif statements are non-standard Fortran, and will break some Fortran compilers.  I would rather just skip the Fortran functions on Windows.  They are not critical to have.  They're just the older Fortran versions of the same functionality that appears in the C functions.  I've kept them for historical reasons.  In the future, I could simply deprecate them or put them in the Demo folder, or perhaps a "Historical" folder, so they'd be available, but not part of the AMD static or dynamic library.

In the meantime, I think it would be safest not to modify them with #if ... #endif, and instead, just not compile them for Windows.

I see that SPQR, GraphBLAS, SPEX, and Mongoose not modified at all by this pull request.   Why is that?  Are they OK as they are?

CCOLAMD has these lines in ccolamd.c:

```
#define GLOBAL CCOLAMD_PUBLIC
#define PUBLIC CCOLAMD_PUBLIC
```

but COLAMD, which is a very similar package, has this in colamd.c:

```
#define GLOBAL
#define PUBLIC
```

From this commit: https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/228/commits/d68338494c624356476cdc6234ccf04b922a190f .  Shouldn't those kind of changes be done for all packages in SuiteSparse?  (Yes, that would be very very tedious to do ...).

Why is UMFPACK and COLAMD different from, say, CCOLAMD?  Do the source code files need any declaration at all on Windows?  I thought they did, which is why all my source files have an internal #define GLOBAL ... which is "extern" except on Windows.  I think the "extern" is optional on Linux, Unix, and the Mac, when the functions are defined in the source.c or source.cpp files.  I thought I had to have them there for Windows.

I need to understand more about how this works in Windows before making these changes, particularly since some of these changes aren't consistent.

Is there some documentation on Windows where I could read up on the __declspec dllimport/dllexport, versus "extern" for static vs dynamic libraries?

DrTimothyAldenDavis:
Regarding the changes to amd.f and amdbar.f:  I've changed the AMD cmake so that it will not build any Fortran at all on Windows, in SuiteSparse v6.0.4.beta1 (just tagged as a prerelease, not yet pushed to the stable branch).

mmuetzel:
> I'm concerned that the #if ... #endif statements are non-standard Fortran, and will break some Fortran compilers. 

The compiler directives start with a comment character. They should look like a comment to any compiler that isn't from the GCC or Intel collection. I needed those directives when I tested with `gfortran`. IIUC, they aren't needed for MSVC and could have been removed with the last commit. I left them in because they don't do any harm.

> I see that SPQR, GraphBLAS, SPEX, and Mongoose not modified at all by this pull request. Why is that? Are they OK as they are?

I can only speak for GCC and Clang compilers in MinGW mode. For them, a library exports all symbols unless there is any symbol that is declared as `dllexport`. In that case, only those symbols that are declared with that attribute are exported from the library. Think of the `visibility("default")` attribute when compiling with hidden visibility on Linux. That's a pretty similar mechanism.
The non-modified libraries didn't use any of those attributes (I guess). So, they didn't need any modifications when I tested with GCC.
It's well possible that more changes are needed so that those libraries export anything with MSVC. (But I don't know tbh.)

> all my source files have an internal #define GLOBAL ... which is "extern" except on Windows.

To be honest, I was a little puzzled when I saw that. An `extern` keyword on a function definition doesn't make any sense to me. It definitely isn't needed on Windows (or Linux afaict). I was assuming that you needed those for some obscure platform or compiler, and just mirrored it for Windows. That's why I didn't touch it in this PR. Unless, it caused compilation errors.
Afaict, `extern` is implicit on function declarations by default ("forward declarations"). So, it isn't needed there either.
It *does* make a difference for variable declarations: Those are static by default afaict.


If the `extern` keyword isn't needed for any platform, it might be best to get rid of it everywhere (except where it is used with variables).

DrTimothyAldenDavis:
Yes, I know that the `extern` statements are not required on Linux or Unix.  The `extern` statements were there because some Windows build system in the past required them.  I can't recall which (it was long before github existed).  So I'd rather get rid of them entirely.

To do that, though, I need to understand exactly when the `__declspec` declarations are required.   I'd prefer to get rid of them entirely.  I see the documentation here: 
https://learn.microsoft.com/en-us/cpp/build/importing-and-exporting?view=msvc-170
and
https://learn.microsoft.com/en-us/cpp/build/dlls-in-visual-cpp?view=msvc-170
but that is a bit opaque to me.

I came across this, which would be a better solution:
https://www.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/

If I can get cmake to create the *.def files then I can avoid all `__declspec` statements entirely.

@eriknw and @jim22k : See the discussion above about __declspec and extern declarations in include files.  The python interface to GraphBLAS uses the GB_PUBLIC statements in the GraphBLAS.h include file.  Is that just a tag used in your preprocessor for cffi?  If so, can it just be the following?

`#define GB_PUBLIC`

mmuetzel:
> I came across this, which would be a better solution:
> https://www.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/

That's very interesting.
I'm currently working on a PR that takes that approach. It will most probably be less intrusive than this PR.
On the other hand, if you ever decide to use visibility flags to only export symbols that are part of the API, you'd probably need to take the approach from here.
If you don't mind that *internal* functions are also being exported from the libraries, the other approach (from your link) is probably the better (because less intrusive) one.

DrTimothyAldenDavis:
I think I will take an entirely different approach to this issue, one based on the discussion at that link at kitware.  I will remove all __declspec from all SuiteSparse packages, at least in the long term.

The only place where this is currently required is the global variable SuiteSparse_config, with function pointers to malloc, calloc, etc.  Global variables require import/export declarations on Windows.

But I can remove that requirement by adding functions like SuiteSparse_config_malloc_func_get(), which simply would do this:

```
void * SuiteSparse_config_malloc_func_get (void)
{
    return ((void *) SuiteSparse_config.malloc_func) ;
}
```

Then the SuiteSparse_config struct, which contains 7 function pointers, could become a local static structure and it wouldn't need any DLL import/export decorations.  I would just have 7 "getter" methods and 7 "setter" methods.

I don't mind using the default visibility of functions, if it allows me to get rid of the __declspec stuff.  I've always disliked those statements and it would be wonderful to remove them from all of SuiteSparse.

I don't think any of my other libraries have any global variables that must be accessible to the user application.

mmuetzel:
> I don't think any of my other libraries have any global variables that must be accessible to the user application.

It's more difficult than I hoped it would be to find those global variable declarations. What about the following ones?
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/dev2/UMFPACK/Source/umf_internal.h#LL739-L740
``` C
GLOBAL int umf_fail, umf_fail_lo, umf_fail_hi ;
GLOBAL int umf_realloc_fail, umf_realloc_lo, umf_realloc_hi ;
```

mmuetzel:
See PR #237 for the alternative approach.

eriknw:
Yeah, I don't think I can add much to this conversation. On the Python side, I think we would just need to try and see.  I'm sure we could process the header file if you set `#define GB_PUBLIC`. It _may_ be nicer to still have `GB_PUBLIC` instead of removing it altogether.

DrTimothyAldenDavis:
Thanks for the info, @eriknw, I will keep GB_PUBLIC for now, and just use "#define GB_PUBLIC" to define it as the empty string.

DrTimothyAldenDavis:
@mmuetzel : please hold off on a new PR on your side.  This update requires some decisions in terms of its effect on the API, and I'd like to handle that.

mmuetzel:
The changes here are probably no longer needed because an alternative approach was used to handle exporting symbols with MSVC.

PR #236, opened 2022-12-30
Linefeed

mmuetzel:
Some (or all?) of the text files that are generated when building are also checked in into the repository.
By default, `cmake` uses `CRLF` (`\r\n`) line endings on Windows. That makes dealing with local checkouts on Windows difficult that are used both for making changes and testing them: A lot of files are differing that don't actually contain changes (apart from the line endings).
Also, Windows is (slowly) migrating away from `CRLF` line endings (see the `*printf` family of functions in UCRT).

This PR explicitly specifies the line ending to be used for generated files to `LF` (which is already the default for `cmake` on Linux - so no changes there).

Alternatively, generated files could be removed from the repository.

Additionally, this PR adds a `.gitattributes` file that specifies that all files should be checked out as is (no encoding translation or line feed changes).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# line endings in repository match line endings on disc
+* -text
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -63,8 +63,10 @@ find_package ( SuiteSparse_config 6.0.2 REQUIRED )
 # configure files
 #-------------------------------------------------------------------------------
 
-configure_file ( "Config/amd.h.in" "${PROJECT_SOURCE_DIR}/Include/amd.h")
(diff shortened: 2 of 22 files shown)


DrTimothyAldenDavis:
Thanks for the update.  I prefer to keep the generated files in the github repo, so adding the NEWLINE_STYLE option is the best solution.

DrTimothyAldenDavis:
Minor update:  I'd like to keep the GraphBLAS/cpu_features package unchanged.  It's a copy from https://github.com/google/cpu_features.  I don't actually use the CMakeLists.txt file from cpu_features to build the library.  It's built via the GraphBLAS cmake itself.  So the changes to GraphBLAS/cpu_features/CMakeLists.txt have no effect.

PR #237, opened 2022-12-30
Export all symbols also with MSVC

mmuetzel:
This is an alternative approach to #228.

Instead of using `dllexport`/`dllimport` attributes, this lets cmake create .def files by setting `WINDOWS_EXPORT_ALL_SYMBOLS ON` for shared libraries. This is working with GCC on Windows for me.
But cmake doesn't seem to create the .def files for that compiler (and it doesn't need to).
I hope it'll do the right thing with MSVC.

I'm not sure if I found all variable declarations that still need to be extern. I only made changes concerning that in SuiteSparse_config and UMFPACK.


DrTimothyAldenDavis:
I will need to handle this update myself, because it involves some changes to global variables and the introduction of some new methods.  The ones in UMFPACK are for debugging only, and I do not need to enable the debug mode on Windows.

SuiteSparse_config has a global struct, but I will make it static.  I'm adding functions like SuiteSparse_config_malloc_func_get for example.

mmuetzel:
> The ones in UMFPACK are for debugging only, and I do not need to enable the debug mode on Windows.

IIUC, they still need to be declared `extern` also for Linux. But maybe you won't need to make them available with the `dllimport` mechanism for MSVC if you disable that feature for that compiler.

DrTimothyAldenDavis:
That's correct, they do need to be extern since they are needed on Linux only.  I'll just use "extern" and not use the __declspec.

DrTimothyAldenDavis:
See my latest commit to the dev2 branch.  The only global variables that need to appear with external visibility are in GraphBLAS (the built-in GrB_Type, GrB_UnaryOps, etc).  I have a new GB_GLOBAL for those objects.

I took out the external visibility of the SuiteSparse_config struct, and added set/get functions to extract or modify its contents.

I think this version should compile on Windows, with either MS Visual Studio or MINGW.  I'll push this to dev to see if the CI works there.  I still have to test this on my Linux desktop ... so far I've only tried it on my Mac.

mmuetzel:
You probably still need to have cmake create and link with .def files for MSVC even after the recent changes.

I rebased on a current head of `dev2` with the remaining parts of this PR.

DrTimothyAldenDavis:
Looks reasonable -- I'll merge those changes and give it a try.

DrTimothyAldenDavis:
I did add this to SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/2d835d24f930b4f1b87b7efbca4a94c1cd72364d/SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake#L86-L88

so the per-target setting of `WINDOWS_EXPORT_ALL_SYMBOLS` might not be necessary.  It's fine having it there as well, just in case, though.

DrTimothyAldenDavis:
The `include ( SuiteSparsePolicy )` comes before any target is defined, in each package.

DrTimothyAldenDavis:
I did a pull request from dev2 to dev, so it will now trigger the CI on these updates.

Thanks for the help with this.  The DLL import/export of all of SuiteSparse is now much cleaner.

PR #246, opened 2023-01-01
CI: Add rules for macos runner

mmuetzel:
This adds rules to run the CI on GitHub-hosted macos runners.

After a bit of fiddling, the CI seems to do something (which is hopefully the right thing) on the runners. I don't have a Mac. So, this is only tested on a feature branch of my clone.

I'm debating whether it would make sense to trigger the CI for a PR. Do you always create PRs when you merge the branches? Or do you also sometimes push a merge directly from your local repository?

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -58,7 +58,7 @@ jobs:
           echo "timestamp=$(date +"%Y-%m-%d_%H-%M-%S")" >> $GITHUB_OUTPUT
 
       - name: restore ccache
-        # setup the github cache used to maintain the ccache from one job to the next
+        # setup the GitHub cache used to maintain the ccache from one job to the next
         uses: actions/cache@v3
         with:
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Thanks!

DrTimothyAldenDavis:
I don't commit changes directly to stable or dev, or at least not usually.  I normally commit to the dev2 branch, and then when I think things look good, I do a PR from dev2 to dev.  Once the CI passes on dev, and once I think the code is ready for the world (as a beta release perhaps), I do a PR from dev to stable.

mmuetzel:
Taking your workflow into account, it might make sense to change the trigger for the CI from `push` to `pull_request`. That way, the CI would still not run for your usual pushes to the `dev2` branch (like you wished). But you could check *before* merging if the CI passes.

mmuetzel:
See #250.

PR #248, opened 2023-01-01
Avoid name conflict of static libraries with MSVC

mmuetzel:
During some early attempts to use MSVC to compile SuiteSparse, I ran into errors when building static libraries is not disabled.

There are three "types" of libraries on Windows:
1. static libraries (like on POSIX)
2. dynamic libraries (mostly comparable to shared libraries on POSIX)
3. import libraries

There is no equivalent for the last one on POSIX. To get a rough idea: All symbols must be resolved on link-time on Windows (similar to linking with `-no-undefined` on POSIX). The import libraries are static libraries that provide "glue" at link time that allows using the symbols exported by the dynamic libraries on runtime. (GCC uses some tricks to allow linking to the dynamic libraries "directly" without import libraries. But for some reason it is "better" to link using import libraries.)

For MinGW the default file extensions for those libraries are:
1. `.a`
2. `.dll`
3. `.dll.a`

MSVC uses the following default file extension:
1. `.lib`
2. `.dll`
3. `.lib`

That means, there is a name collision for the static and import libraries if they use the same names. And indeed, ninja fails with an error that multiple rules exist for the same target.

This PR proposes changes to avoid that name conflict by renaming the static libraries for MSVC.

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -104,6 +104,11 @@ if ( NOT NSTATIC )
         C_STANDARD_REQUIRED 11
         OUTPUT_NAME amd
         SOVERSION ${AMD_VERSION_MAJOR} )
+
+    if ( MSVC )
+        set_target_properties ( amd_static PROPERTIES
+            OUTPUT_NAME amd_static )
+    endif ( )
 endif ( )
 
 #-------------------------------------------------------------------------------
--- a/BTF/CMakeLists.txt
+++ b/BTF/CMakeLists.txt
@@ -79,13 +79,18 @@ set_target_properties ( btf PROPERTIES
 #-------------------------------------------------------------------------------
 
 if ( NOT NSTATIC )
-add_library ( btf_static STATIC ${BTF_SOURCES} )
+    add_library ( btf_static STATIC ${BTF_SOURCES} )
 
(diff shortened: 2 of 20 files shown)


DrTimothyAldenDavis:
Thanks for the update.  I'm in the middle of modifying some of the CMakeLists.txt files though, to fix the static linking problem on Windows. We might need to rebase your PR once I finish this update.  Most of my changes are in the Find*.cmake files, however.  I'll post my updates shortly and then try to merge your PR.

DrTimothyAldenDavis (approved):

DrTimothyAldenDavis:
How does this affect the Find*.cmake scripts?  The FindAMD.cmake script for example has this, to find the static library:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/2820df2e0856a5a7580e18963ac9516a10c72d23/AMD/cmake_modules/FindAMD.cmake#L57-L66

Note the line "NAMES amd".  Will that need to be "NAMES amd_static" on MSVC?

DrTimothyAldenDavis:
I've merged your changes and mine to the dev2 branch, and I just did a PR from dev2 to dev, where the CI is testing everything now.

mmuetzel:
> How does this affect the Find*.cmake scripts?

You are right. Those need to be changed. See #251.

PR #251, opened 2023-01-02
Fix detecting static libraries with MSVC

mmuetzel:
Static libraries need to use a different name than shared libraries for MSVC (see #248).
At first, I consider to just prefer static libraries with the specific name. But that could have lead to a falls detection of an import library as a static library with MSVC.

This PR changes the rules to *only* look for the names specific to the respective compiler.


DrTimothyAldenDavis:
Looks great.  I see you also found a bug in FindSPQR_CUDA.cmake, where I was using the wrong name for static library.  Thanks for catching that.

PR #253, opened 2023-01-02
Fix typo in restore-keys identifier.

mmuetzel:
There is a typo in the identifier for the restore keys. Sorry for that.


mmuetzel:
It still didn't pick up the cache. Hmm... 🤔

mmuetzel:
It looks like caches cannot be picked up from a different branch (but from the default branch):
https://github.com/orgs/community/discussions/27059#discussioncomment-3254477

What a strange restriction...

Edit: With the last changeset, it should pick up the last cache from the stable branch (the current default branch). That could still be useful if the branches didn't diverge too much...

DrTimothyAldenDavis:
Thanks for the update.

Issue #234, opened 2022-12-29
Porting the new cmake-based build system to MSVC


Fabian188:
Compiling 6.0.3 in Win10 with MSVC fails for CCOLAMD with the error

 16%] Building C object CMakeFiles/ccolamd.dir/Source/ccolamd.c.obj
...\CCOLAMD\Source\ccolamd.c(1080): error C2491: "ccolamd_recommended": definition of dllimport function not allowed
...\CCOLAMD\Source\ccolamd.c(1109): error C2491: "ccolamd_set_defaults": definition of dllimport function not allowed
...\CCOLAMD\Source\ccolamd.c(1150): error C2491: "csymamd": definition of dllimport function not allowed
...\CCOLAMD\Source\ccolamd.c(1478): error C2491: "ccolamd": definition of dllimport function not allowed
...\CCOLAMD\Source\ccolamd.c(1519): error C2491: "ccolamd2": definition of dllimport function not allowed
...\CCOLAMD\Source\ccolamd.c(1965): error C2491: "ccolamd_report": definition of dllimport function not allowed
...\CCOLAMD\Source\ccolamd.c(1978): error C2491: "csymamd_report": definition of dllimport function not allowed
...

**Describe the bug**

**To Reproduce**
In a clean 6.0.3 I do the following steps after setvars to initialize Intel oneAPI
cd Suitesparse_config\build
cmake .. -G "NMake Makefiles"
nmake

repeat for AMD and CAMD and all works
for CCOLAMD I get the above error 

This is the configure vor CCOLAMD

```
-- Building CCOLAMD version: v3.0.2 (Dec 9, 2022)
-- Source:        C:/Users/fabia/code/master/release_msvc/cfsdeps/suitesparse/src/suitesparse/CCOLAMD 
-- Build:         C:/Users/fabia/code/master/release_msvc/cfsdeps/suitesparse/src/suitesparse/CCOLAMD/build
-- The C compiler identification is MSVC 19.34.31937.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.34.31933/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Install rpath:  
[...]

Issue #175, opened 2022-11-10
Suggestion for root CMakeLists.txt


Fabian188:
I propose a simple CMakeLists.txt

It makes use of the SuiteSparse CMake files via cmake external projects and just adds the root CMakeLists.txt.

This is just a template with a small selection of options to demonstrate how it could be re realized.

I'm far from a CMake guru but it appears as an appropriate way for a simple wrapper for the original SuiteSparse CMake structure.

Please feel free to use and modify the file as you like. I wrote it myself and as it is so small and easy I put it into public domain without any restrictions - no need to mention me. 

Example usage for an empty SuiteSparse folder: mkdir mybuild; cd mybuild; ccmake ..

When changing the CMakeLists.txt for some major changes it is necessary to clean the build directory (here mybuild) and call make clean of SuiteSparse. I testet it with dev2 from Nov 10 

[CMakeLists.txt](https://github.com/DrTimothyAldenDavis/SuiteSparse/files/9985263/CMakeLists.txt)

DrTimothyAldenDavis:
Thanks -- I'll take a look.

fabiencastan:
@Fabian188 Very useful! It works fine for me.
I just needed to remove all the lines with:
`INSTALL_COMMAND "" # if removed, tried to call make install in build`
If we set these lines, it disables the installation of the libraries, which is quite problematic!
If we remove these lines and run cmake with `cmake -DCMAKE_INSTALL_PREFIX=/custom/path`, it installs everything correctly in the custom path.

fabiencastan:
Making a confusion in my various tests, I also needed to propagate the CMAKE_INSTALL_PREFIX:
```
set(CMAKE_ARGS
    -DNSTATIC:BOOL=(NOT BUILD_STATIC)
    -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}  # <<< added this line
    )
```

Fabian188:
@fabiencastan thanks for the input. I just created a fork SuiteSparse-root-cmake and included your suggestion.

Feel free to suggest more :)

Fabian188:
What I do not like with this approach is, that a "make" implies a "make install". This is probably due to the external projects.

fabiencastan:
@Fabian188 Do you have a link to your proposal? Or could you create a draft PR?

fabiencastan:
The link: https://github.com/Fabian188/SuiteSparse-root-cmake

[...]

PR #255, opened 2023-01-02
SuiteSparse v7.0.0.beta1

DrTimothyAldenDavis:
Jan 1, 2023: SuiteSparse version 7.0.0.beta1

    * SuiteSparse_config: now v7.0.0
    * SuiteSparse_config struct: removed from external visibility to simplify
        the Windows build, so that no global data is externally visible.
        This requires a major version number increase from v6.x to v7.x for the
        SuiteSparse meta-package (which has the same version number as
        SuiteSparse_config).  Added get/set methods to SuiteSparse_config to
        access the contents of the struct.
    * NFORTRAN: option added to ignore any Fortran methods, even if a Fortran
        compiler is available.
    * port of new cmake-based build system to Windows
    * NFORTRAN: option added to disable Fortran entirely
    * GraphBLAS v7.4.1:  global free pool disabled, and GrB_mxm heuristics
        revised.

Most of the changes are for the port of the new cmake build system to Windows, using various build systems (MS Visual Studio, MSYS2, and lots of different compilers).  GraphBLAS v7.4.1 has some performance enhancements as well, with a revision of the heuristics for GrB_mxm.


h-vetinari:
Hello! Thank you very much for your continued very active maintenance of this project!

I was very happy to see that basically all subprojects have a `CMakeLists.txt` now, and was wondering if you'd be willing to add a top-level `CMakeLists.txt` that allows building all of SuiteSparse at once (equivalently to the `Makefile`).

I'm happy to open an issue for this if more appropriate, but given that this PR was open I thought it might be worth commenting for such a major version bump.

Obviously you reference https://github.com/jlblancoc/suitesparse-metis-for-windows on your website already - presumably the `CMakeLists.txt` [there](https://github.com/jlblancoc/suitesparse-metis-for-windows/blob/master/CMakeLists.txt) would already have most of the necessary bits for such a toplevel build file.

h-vetinari:
PS. I saw the related https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/234. Once there's a git tag, I'd be happy to test 7.0.0.beta1 on conda-forge infrastructure on windows. This was actually one of my main reasons for thinking about a top-level build-file, then I wouldn't have to go build each subfolder separately (but can be done if necessary).

PPS. I have (unfortunately 😅) some experience with the kind of missing symbol problems described in #234. I might be able to provide a patch (no promises of course!).

DrTimothyAldenDavis:
Fabien Wein (@Fabian188) has proposed a top-level CMakeLists.txt, here: https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/175 .  Once the current version on the dev2 branch is working correctly, I'd like to consider that.  The CMakeLists.txt file posted there (not as a pull request, but just as a single file) would likely need to be updated with all the recent changes.

I can't use a mechanism that requires a top-level CMakeLists.txt or that makes deep changes to each of the packages.  Each package has to be stand-alone and needs to be able to work without the top-level CMakeLists.txt.  I'm open to the idea of a single top-level CMakeLists.txt that can take the place of my top-level Makefile, though.  I think that would simplify things for Windows users.

DrTimothyAldenDavis:
The structure at https://github.com/jlblancoc/suitesparse-metis-for-windows won't work with the current version of SuiteSparse.  That's a monolithic system that builds all packages and requires the top-level CMakeLIsts.txt.  I can't do that.

DrTimothyAldenDavis:
And the help will be greatly appreciated -- thanks!  I don't do any development on Windows, myself, and there are parts of the MS Visual Studio build system that are puzzling to me.

Fabian188:
@DrTimothyAldenDavis  The file I once posted is obsolete, the current version is https://github.com/Fabian188/SuiteSparse-root-cmake
The two files are root CMakeLists.txt and suitesparse_install.cmake.in.

It is meant to be configured by ccmake, builds only CHOLMOD and UMFPACK up to now and is probably a complete other coding style than yours but please feel free do adopt it to your needs. I think it could help usage and make the manual Makefiles obsolete. I believe usage of cmake external projects is the best way to integrate independent cmake projects.

h-vetinari:
> I can't use a mechanism that requires a top-level CMakeLists.txt or that makes deep changes to each of the packages. Each package has to be stand-alone and needs to be able to work without the top-level CMakeLists.txt.

Yeah, I understood that afterwards, and it makes a lot of sense.

I'll try to get things running on our side (with or without a top-level file) as soon as there's a tag.

> I believe usage of cmake external projects is the best way to integrate independent cmake projects.

This might be fine for personal use but it's quite an antipattern to always fetch and rebuild other libraries (especially for distributors and integrators, where that approach would lead to gratuitous duplication and symbol clashes). Those projects should be built separately, be present in whatever `CMAKE_INSTALL_PREFIX` is, and then set up with `find_package`.

Fabian188:
@h-vetinari:
> This might be fine for personal use but it's quite an antipattern to always fetch and rebuild other libraries (especially for distributors and integrators, where that approach would lead to gratuitous duplication and symbol clashes). Those projects should be built separately, be present in whatever `CMAKE_INSTALL_PREFIX` is, and then set up with `find_package`.

I believe, there is a misunderstanding. SuiteSparse consists of several independent cmake projects and these are expected to be configured and built in their respective build orders. There are handwritten Makefiles which automate this. My proposed CMakeLists.txt does exactly this: generate and build the cmake projects in their build directories. Nothing more and nothing less, therefore I see no duplication and symbol clashes. It is just a more convenient and portable way via cmake/ccmake than the Makefiles. The same major work is in both case done by the project's own cmake files.

h-vetinari:
> SuiteSparse consists of several independent cmake projects and these are expected to be configured and built in their respective build orders.

Admittedly I don't know the project in detail, but from a high-level POV, that build-order coordination is only in-scope for a top-level `CMakeLists.txt`. If you want to build only a given subproject, it's IMO the user's responsibility that the required preconditions are met (including the building of other subprojects).

Basically, from a separation of concerns POV, building one subproject should not trigger the building of other subprojects. Not sure if such a split is realistic or desirable for suitesparse, but it would be much cleaner, and still pretty easy IMO (e.g. if `find_package(suitsparse::<other_subproject> REQUIRED)` fails, you'll get a pretty self-explanatory error message).

h-vetinari:
Or, as always in CMake, you can have your cake and eat it too, provided you're willing to have the complexity of a `SUITESPARSE_BUILD_ISOLATION` (or whatever) flag + switches everywhere.

Fabian188:
@h-vetinari It would help not to speak from a general POV but from the specific implementation.

Are you concerned that e.g. always SuiteSparse_config is built, also AMD, CAMD, COLAMD, and CCOLAMD needs not to be disabled when one wants CHOLMOD? That could be done different. But if it is only CHOLMOD you want wth the  SuiteSparse_config, AMD, CAMD, COLAMD, and CCOLAMD from your system one can directly call the CMakeLists.txt from CHOLMOD.

PR #258, opened 2023-01-04
SPEX: Don't override CMAKE_PREFIX_PATH defined as command line argument

mmuetzel:
After recent changes, linking was prefering static libraries even for shared libraries on Windows.

Change the `Find*.cmake` modules to prefer linking to import libraries on Windows. Fall back to linking to the shared library directly unless MSVC.

### DIFF
--- a/SPEX/cmake_modules/FindGMP.cmake
+++ b/SPEX/cmake_modules/FindGMP.cmake
@@ -17,8 +17,6 @@
 # GMP_LIBRARIES   - libraries when using gmp
 # GMP_FOUND       - true if gmp found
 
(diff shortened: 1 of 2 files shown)


mmuetzel:
Hold on. Maybe most of this isn't needed at all. I was basing this on an older head.

The only part that might be useful is part of the changes for the GMP and MPFR modules. (Don't override passed arguments with environment variable.)

mmuetzel:
Force pushed to only contain the actually useful changes.

DrTimothyAldenDavis:
Thanks!  I'll merge it in and we can try to get this all to work ...

PR #259, opened 2023-01-04
Look for libraries in folders for MSVC configurations

mmuetzel:
MSVC apparently has two different configurations for compilation: Release and Debug.
For each of these configurations, it creates the binaries in different subfolders of the build folder.

Adjust the `Find*.cmake` modules to detect the libraries in the corresponding subfolders.


DrTimothyAldenDavis:
Looks great -- I will merge it in.

PR #260, opened 2023-01-04
Add CI rules for MSVC

mmuetzel:
Add rules to build (most of) the SuiteSparse libraries in CI with MSVC.

This is currently failing and depends on #259 (and possibly #258).

It currently only builds the libraries and doesn't run the demos to check if they actually work. But this is still a good enough first step imho.

Currently, CXSparse doesn't build because there are no `complex` floating point type with MSVC. Afaict, those types are optional as of C11. And MSVC was never C99-compatible. See also:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/complex-math-support?view=msvc-170
https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/#whats-not

Additionally, GraphBLAS doesn't build with the `Ninja*` generators. (See https://github.com/DrTimothyAldenDavis/GraphBLAS/issues/194.)
I does build with the `Visual Studio *` generators. But those don't support `ccache`. I figured having a working compiler cache is preferable for CI that might be running often.

PS: I added the manual loop over all libraries because MSVC can't use the top-level Makefile. (Or I didn't figure out how to do that.)
But come to think of it, I prefer having the option to insert action "directives" because it makes looking for the output of a given library so much easier. Should we change the other rules to also not use the top-level Makefile?

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -279,3 +279,169 @@ jobs:
       - name: ccache status
         continue-on-error: true
         run: ccache -s
+
+
+  msvc:
+    # For available GitHub-hosted runners, see:
+    # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
+    runs-on: windows-latest
+
+    defaults:
+      run:
+        # Use bash as default shell
+        shell: bash -el {0}
+
+    env:
+      CHERE_INVOKING: 1
+
+    steps:
+      - name: get CPU name
+        shell: pwsh
+        run : |
+          Get-CIMInstance -Class Win32_Processor | Select-Object -Property Name
+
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
I'm OK merging it in now, before it's working, if that's easier.  Or we can wait until it works.  Either way is fine with me.

mmuetzel:
I'll rebase this on the current `dev2` now that the other PRs are merged.
That way, we can check if this PR is ok or if I missed some parts...

mmuetzel:
It builds with a bunch of warnings. But it builds. 🎉

Good to go from my side. 👍

DrTimothyAldenDavis:
OK -- I'll merge it in -- thanks!

PR #262, opened 2023-01-04
Combine output in groups instead of using top-level Makefile in CI

mmuetzel:
Instead of using the top-level Makefile, combine the output in groups for all platforms. That makes it easier to find the log for a given library.

For some reason, mingw (CLANG64) started to fail its tests. I'm not quite sure why.
~~Maybe because of this?~~
~~https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/9f678372a8bf7d956cfe20df55af6bfadbd146e8#diff-6dbf8025b96177a872c950c1dc048aab951ff1d97fda27356fa6f723367b312dL39~~

~~Would that module still need to be included when configured to build the demos?~~

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,6 +9,13 @@ on:
 
 concurrency: ci-${{ github.ref }}
 
+env:
+  # string with name of libraries to be built
+  BUILD_LIBS: "SuiteSparse_config:Mongoose:AMD:BTF:CAMD:CCOLAMD:COLAMD:CHOLMOD:CSparse:CXSparse:LDL:KLU:UMFPACK:RBio:SuiteSparse_GPURuntime:GPUQREngine:SPQR:GraphBLAS:SPEX"
+  # string with name of libraries to be checked
+  CHECK_LIBS: "SuiteSparse_config:Mongoose:AMD:BTF:CAMD:CCOLAMD:COLAMD:CHOLMOD:CSparse:CXSparse:LDL:KLU:UMFPACK:RBio:SPQR:GraphBLAS:SPEX"
+
+
 jobs:
 
   ubuntu:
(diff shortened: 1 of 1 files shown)


mmuetzel:
The build error for mingw (CLANG64) is the following:
```
  [39/44] Linking Fortran executable umf4zhb.exe
  FAILED: umf4zhb.exe 
  cmd.exe /C "cd . && C:\msys64\clang64\bin\flang.exe   CMakeFiles/umf4zhb.dir/Demo/umf4zhb.f.obj CMakeFiles/umf4zhb.dir/Demo/umf4_f77zwrapper.c.obj -o umf4zhb.exe -Wl,--out-implib,libumf4zhb.dll.a -Wl,--major-image-version,0,--minor-image-version,0  libumfpack.dll.a  D:/a/SuiteSparse/SuiteSparse/SuiteSparse_config/build/libsuitesparseconfig.dll.a  C:/msys64/clang64/lib/libomp.dll.a  D:/a/SuiteSparse/SuiteSparse/AMD/build/libamd.dll.a  C:/msys64/clang64/lib/libopenblas.dll.a  D:/a/SuiteSparse/SuiteSparse/CHOLMOD/build/libcholmod.dll.a  D:/a/SuiteSparse/SuiteSparse/CHOLMOD/build/GPU/libcholmod_cuda.dll.a  D:/a/SuiteSparse/SuiteSparse/AMD/build/libamd.dll.a  C:/msys64/clang64/lib/libopenblas.dll.a  D:/a/SuiteSparse/SuiteSparse/CHOLMOD/build/libcholmod.dll.a  D:/a/SuiteSparse/SuiteSparse/CHOLMOD/build/GPU/libcholmod_cuda.dll.a  D:/a/SuiteSparse/SuiteSparse/COLAMD/build/libcolamd.dll.a  D:/a/SuiteSparse/SuiteSparse/CAMD/build/libcamd.dll.a  D:/a/SuiteSparse/SuiteSparse/CCOLAMD/build/libccolamd.dll.a && cd ."
  ld.lld: error: undefined symbol: imag_
  >>> referenced by D:/a/SuiteSparse/SuiteSparse/UMFPACK/build/./CMakeFiles/umf4zhb.dir/Demo/umf4zhb.f-pp.f:53
  >>>               CMakeFiles/umf4zhb.dir/Demo/umf4zhb.f.obj:(_QQmain)
  >>> referenced by D:/a/SuiteSparse/SuiteSparse/UMFPACK/build/./CMakeFiles/umf4zhb.dir/Demo/umf4zhb.f-pp.f:79
  
  >>>               CMakeFiles/umf4zhb.dir/Demo/umf4zhb.f.obj:(_QQmain)
  
  flang: error: linker command failed with exit code 1 (use -v to see invocation)
```
Any clue as to why that is failing?

DrTimothyAldenDavis:
Not sure why imag(x) should suddenly fail.  I've enabled Fortran.  I moved the content of SuiteSparseFortran.cmake back into SuiteSparsePolicy.cmake, because of a linking error on a Linux system with both GNU opemp, Intel openmp, and CUDA.  Something in the sequence was causing a failure.  So the SuiteSparseFortran.cmake has no effect here.

The imag(x) returns the imaginary part of a complex number x in Fortran.  That function has been there forever.

Maybe it should be aimag(x), but that seems to return a single precision result.   I need the double precision result.

See https://gcc.gnu.org/onlinedocs/gfortran/AIMAG.html

mmuetzel:
I just realized that the order in the top-level Makefile for the demos is different from the order for the build commands.
I used the same order for both. Could that be the reason?

DrTimothyAldenDavis:
I doubt it.  I think the standard Fortran 77 way is to use real(z) and aimag(z) to get the real and imaginary parts of a complex number, with the result being in the same precision as the input.  I just pushed a change to the umf4zhb*.f demos.

mmuetzel:
I was under the wrong impression that this was a new error. But it didn't compile also before this change. 
See, e.g.: https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/3839164607/jobs/6536564219#step:9:2557

I'll rebase on top of your changes.

mmuetzel:
Oof. That might be a limitation of the current state of LLVM Flang:
https://discourse.llvm.org/t/request-advice-on-reporting-many-errors-found-in-fortran-test-sets/65868/8

Maybe, we should disable Fortran for the demos for the CLANG64 environment until LLVM 16 will be released and packaged by MSYS2...

But let's see what your changes did. After all, `aimag` is not in the list in that forum thread.

DrTimothyAldenDavis:
Things seem to be working.  Once it finishes, I will pull all the logs for this run of the CI, and go through any warnings and try to fix them.

Is there a way to force a clean rebuild of everything, so I can check the entire build for warning messages?  I could probably do that by revising the date of each package (they are all tagged Jan 1, 2023, and I'll want to fix that date anyway).

mmuetzel:
> Is there a way to force a clean rebuild of everything, so I can check the entire build for warning messages? I could probably do that by revising the date of each package (they are all tagged Jan 1, 2023, and I'll want to fix that date anyway).

What do you mean by "clean rebuild"? The CI always builds from a freshly cloned repository.

DrTimothyAldenDavis:
I've noticed that the CI runs are very fast if I make a small change.  That's the cache I assume?

mmuetzel:
> I've noticed that the CI runs are very fast if I make a small change. That's the cache I assume?

Yes, that should be the effect of using ccache.

PR #264, opened 2023-01-05
Build GraphBLAS with MSVC in CI

mmuetzel:
GraphBLAS can be built with MSVC by the CI now.
Also increase the ccache size to account for this change.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -445,7 +445,7 @@ jobs:
         # Limit the maximum size and switch on compression to avoid exceeding the total disk or cache quota.
         run: |
           test -d ${{ steps.ccache-cache.outputs.ccachedir }} || mkdir -p ${{ steps.ccache-cache.outputs.ccachedir }}
-          echo "max_size = 50M" > ${{ steps.ccache-cache.outputs.ccachedir }}/ccache.conf
+          echo "max_size = 250M" > ${{ steps.ccache-cache.outputs.ccachedir }}/ccache.conf
           echo "compression = true" >> ${{ steps.ccache-cache.outputs.ccachedir }}/ccache.conf
           ${CCACHE} -p
           ${CCACHE} -s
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Also: CXSparse should build on Windows now.  It has always had a -DNCOMPLEX=1 option to disable the complex data type, but the cmake was not setting this flag.  I changed the CXSparse/CMakeLists.txt to set -DNCOMPLEX=1 for MSVC.

mmuetzel:
I tried with `-DNCOMPLEX=ON` for CXSparse a couple days ago. Then, it didn't work. I'll try again now that compilation of a couple of files is excluded with your recent changes...

DrTimothyAldenDavis:
With -DNCOMPLEX=1, CXsparse also needed be compiled without the `cs_ci_*.c` and `cs_cl_*.c` source files, and a few others.  I added that revision to the CXSparse/CMakeLists.txt

DrTimothyAldenDavis:
Looks great.  I've merged it in and also did a PR from dev2 to dev.  CXSparse should be fine with this update; I tested it on my mac with forcing NCOMPLEX on, and it worked fine.  I skipped running the complex demos (see the CXSparse/Makefile where I split the demos into rdemos and cdemos).

mmuetzel:
Thanks for merging.

It might be worth considering including the demos as CTests in the cmake rules. See, e.g.: https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html

That way the cmake configuration could chose which tests to run. And it would allow running the tests on platforms that don't have `make` (e.g., MSVC).
But that is probably a different project for later.

DrTimothyAldenDavis:
Yes, that would be a good idea.  I have another project in the GraphBLAS.org that uses ctest in the build.yaml.  It also does test coverage: https://github.com/GraphBLAS/LAGraph .  The script then posts the test coverage results here:  https://graphblas.org/LAGraph/ .  I didn't write the workflow scripts or the cmake scripts that use ctest.  They're pretty handy though.

PR #267, opened 2023-01-06
Generate cmake modules automatically

mmuetzel:
This is a proposal to let cmake generate the modules that are needed to use the libraries from SuiteSparse automatically.

Adjusting the `Find*.cmake` modules to have them detect the right thing on any platform has proven to be difficult. And the current implementation might not work correctly if cmake was configured in an unexpected way. The automatically generated files have the advantage that we don't have to worry about all possible configurations, but we can rely on cmake doing "The Right Thing"™.

In the long run, the automatically generated modules might be able to replace the current `Find*.cmake` files entirely.

The changes in this PR only touch `SuiteSparse_config`. They aren't meant to be merged (yet). But I hope they can illustrate how those modules could be generated in a "more default" way.

To test, run the following commands in the `build` directory of `SuiteSparse_config`:
```
cmake -DCMAKE_INSTALL_PREFIX=../.. ..
cmake --build .
cmake --install .
```
The generated .cmake modules should be installed in `lib/cmake/SuiteSparse_config` in the top level of the repository. It should be possible to use them with `find_package` (if the full path to `lib/cmake` is added to `CMAKE_MODULE_PATH`).


mmuetzel:
I added another commit so that the generated modules set the same variables that the `Find*.cmake` module currently does.
But it might be easier if dependent projects would just link to the imported `suitesparseconfig` or `suitesparse_static` targets instead.

mmuetzel:
For a test, I used the following `CMakeLists.txt` that I saved to `SuiteSparse_config/test_cmake`:
```
cmake_minimum_required ( VERSION 3.22 )

project ( test_cmake )

find_package ( SuiteSparse_config 7.0.0 )
```
And made sure the module can be found with `cmake -DCMAKE_MODULE_PATH=../../lib/cmake .`.

mmuetzel:
The last change changes the other projects to use the imported autogenerated target instead of the `FindSuiteSparse_config.cmake` module.
Untested as of now and will likely fail...

mmuetzel:
Seems to be working now.

Generating and using the exported targets is pretty easy. The most complicated part was getting a working fallback mechanism for the variables that are defined by the Find*.cmake module.

For the motivation of the whole thing: I was looking into how to setup ctests. That is a bit problematic with the current modules: The libraries aren't found on runtime on Windows.
If I understand the documentation correctly (and I'm not sure if I do), the targets contain information as to where the libraries are installed. And ctest might be able to utilize that information. With the current approach (variable set by the Find*.cmake modules), that information is not available to other projects.

mmuetzel:
Is the approach with the exported cmake targets something you'd consider accepting?
It has the additional advantage that it would no longer use heuristics to locate headers and libraries. Instead it would reliably point to the actually installed headers and libraries.

In case you'd accept this, I could look into adapting the other libraries, too. If you don't, I won't waste my time on it. 😀

DrTimothyAldenDavis:
Yes, I'm interested in considering it.  I was just reading over the draft pull request now, before your last comment above, in fact.

Since the changes to the build system look significant, I should probably try them out in your fork of SuiteSparse in your cmake branch (  https://github.com/mmuetzel/SuiteSparse/tree/cmake ).  Let me give that a try before you go deeper with all the other pacakges.

Does this method generate the FindSuiteSparse_config.cmake script?

What do you think about the timing of adding this feature?  I'd like to post SuiteSparse v7.0.0.beta1 at some point soon.  Maybe this could be in a subsequent release, like v7.1.0?  If we do that, I could move any additional changes I can't make right away into a v7.2.0 soon thereafter.

DrTimothyAldenDavis:
Would this work with a top-level CMakeLists.txt if I were to add one?

See https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/175

mmuetzel:
> Does this method generate the FindSuiteSparse_config.cmake script?

It doesn't generate that file. But it autogenerates a few target files. To a user, they should behave just the same. I tried to have it define all variables that are currently defined in that file. But instead of using those, it would be easier for an end user to just link to the imported target.

> What do you think about the timing of adding this feature? I'd like to post SuiteSparse v7.0.0.beta1 at some point soon. Maybe this could be in a subsequent release, like v7.1.0? If we do that, I could move any additional changes I can't make right away into a v7.2.0 soon thereafter.

I currently don't know at which pace I will be able to work on this during the next couple of weeks. It might take some time before this will be finished. But all the heavy lifting should be done and the rest is hopefully just some copy/paste and modification for each package.

> Would this work with a top-level CMakeLists.txt if I were to add one?

I'd guess that it wouldn't matter. The generated files export information about the targets. That information should still be valid with our without a top-level CMakeLists.txt.

DrTimothyAldenDavis:
OK.  Let me try out the changes related to SuiteSparse_config in your draft branch first, hopefully by today or tomorrow, before you do the rest of the packages.

mmuetzel:
I consolidated the commits and rebased on a current head.

mmuetzel:
This is almost ready for review now. All modules are there.
But there might be an issue for `-DGLOBAL_INSTALL=1 -DLOCAL_INSTALL=1`.

The setup with the global and local installation doesn't look very "`cmake`-y" to me. Usually a user might be expected to configure with `-DCMAKE_INSTALL_PREFIX=../..` if they'd like to have the subprojects installed in the top-level source tree.
What was the motivation for that setup? Would it make sense to remove the `local` and `both` targets from the Makefiles (and remove the `LOCAL_INSTALL` branches from the cmake files)?

DrTimothyAldenDavis:
I need the local install option for various things, such as the simple Makefiles in the Package/Tcov folders.  I also need them for the MATLAB interfaces, which are built inside the various Package/MATLAB folders.  So they're important to keep.

I'm close to getting all the codes ready for a v7.0.0.beta1 release.  Shall I go head and do that now?  I haven't had a chance to review these changes yet.  We can add your changes to the cmake configuraion in v7.0.0.beta2, or perhaps v7.0.1 after I post a stable v7.0.0.  Would that work?

mmuetzel:
> Shall I go head and do that now?

The changes from this PR should definitely wait until after the release. They will probably need some tuning afterwards. 
At the moment, I think there will be some overlinking of shared libraries. (But that might not be a regression from the changes here.) With the targets, it should be quite straightforward to sort that out. But probably better in a follow up. This PR is probably large enough as it is.

> I need the local install option for various things

Thanks for clearing that up for me. Is the `both` target also needed? I.e., would it be possible to have the local and global installation as mutually exclusive alternatives? It would still be possible to install in both prefixes sequentially without needing to rebuild anything. Just not with a single command...

DrTimothyAldenDavis:
Yes, that would be fine to have 2 separate commands and to drop the "make both".

DrTimothyAldenDavis:
I will go ahead and post SuiteSparse v7.0.0.beta1 sometime later today.  It will be a beta release, not a stable one, so I can get feedback from others, after they have a change to try it out.

DrTimothyAldenDavis:
For testing the find_package scripts, I have a simple package, SuiteSparse/Example, that tries to find all packages in SuiteSparse.  It's not built by default.  It assumes all of the packages have been properly installed.  The program is very simple: just calling a single method from each of the packages.  It's a useful way to test all of the find_package methods.

waebbl:
Just looked at the SuiteSparse_config changes.
If I read your changes correctly, you are mixing modern CONFIG style code with outdated MODULE style `find_package_handle_standard_args` calls, i.e. you are using the functions from CMakePackageHelper to transform files in `cmake_modules` which use `find_package_handle_standard_args`. As far as I understand CMake you usually stick to one, nowadays mostly the CONFIG style and don't need any `Find*.cmake` files or the `.cmake` files you provide in the cmake_modules subdirs, as well as the variables which used be in those files. CMake handles most, if not all of them automatically. If a CMakeLists.txt file really needs one of these variables, most time they are extracted with get_target_property from within the CMakeLists.txt file.
Not sure if that's a stable way on how to do this.

mmuetzel:
Thank you for having a look at these changes.
My intention was to make the transition for downstream projects as easy as possible and define all variables that have been defined by the previous modules. To really make sure *all* variables are defined, I figured it wouldn't hurt to call `find_package_handle_standard_args`.
Please, let me know if you think that could be detrimental.

Edit: If we don't care about backward compatibility, the new Config.cmake files could be simplified a lot. Basically everything below the line `# The following is only for backward compatibility with FindSuiteSparse_config.` could potentially be removed.

mmuetzel:
> don't need any `Find*.cmake` files or the `.cmake` files you provide in the cmake_modules subdirs

Could you please point out which of those files are no longer needed. Maybe, they can be removed (in this or another PR).

waebbl:
> Thank you for having a look at these changes. My intention was to make the transition for downstream projects as easy as possible and define all variables that have been defined by the previous modules. To really make sure _all_ variables are defined, I figured it wouldn't hurt to call `find_package_handle_standard_args`. Please, let me know if you think that could be detrimental.

I think it could be detrimental, but I'm not into the cmake sources tbh, so I'm not sure about it. 
Is it out of assumption, some downstream packages need these variables, or do you have an example that illustrates the needs?
In cmake config files there's the possibility to use the `set_and_check` macro to populate variables needed by downstream projects. It can be used for path variables and from the `*.cmake.in` file. In case you are not aware, there's some documentation at [1][2] and the `find_package` documentation. Could this macro be used to fill in the variables and avoid the `find_package_handle_standard_args`?
Instead of printing a status message and some variables after each `find_package_handle_standard_args` in each config file, we _could_ also use the `FeatureSummary`[3] module to print a global summary at the end of the toplevel CMakeLists.txt file and populate it with whatever information is deemed relevant. This way, all information is in one place, and not scattered throughout the cmake output.

> 
> Edit: If we don't care about backward compatibility, the new Config.cmake files could be simplified a lot. Basically everything below the line `# The following is only for backward compatibility with FindSuiteSparse_config.` could potentially be removed.

The cmake build system is quite new to the package, so I wouldn't argue for dropping backward compatibility at this time. IMO this could only be thought about, once the build system has been matured.



> > don't need any `Find*.cmake` files or the `.cmake` files you provide in the cmake_modules subdirs
> 
> Could you please point out which of those files are no longer needed. Maybe, they can be removed (in this or another PR).

Sorry, my bad. I just saw the `Find*.cmake` files in the list of changed files, but didn't notice that you actually removed them. If we provide the `<package>Config{,Version}.cmake` and resp. `*Targets{,-<BUILD_TYPE>}.cmake` we won't need any package internal `Find*.cmake` files, like e.g. `FindAMD.cmake`, which you all dropped.

[1] https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#using-pre-built-packages-with-find-package
[2] https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html
[3] https://cmake.org/cmake/help/latest/module/FeatureSummary.html#module:FeatureSummary

mmuetzel:
> do you have an example that illustrates the needs?

Do you have a list of all projects that use libraries from SuiteSparse?

> Could this macro be used to fill in the variables and avoid the `find_package_handle_standard_args`?

Do you know which variables `find_package_handle_standard_args` sets that we might need to set manually?

> Instead of printing a status message and some variables after each `find_package_handle_standard_args` in each config file, we _could_ also use the `FeatureSummary`[3] module to print a global summary at the end of the toplevel CMakeLists.txt file and populate it with whatever information is deemed relevant. This way, all information is in one place, and not scattered throughout the cmake output.

That sounds like a potentially nice idea for a different PR.

> The cmake build system is quite new to the package, so I wouldn't argue for dropping backward compatibility at this time. IMO this could only be thought about, once the build system has been matured.

Sorry. I don't get your point. Are you arguing for dropping backward compatibility or against dropping it?
Let's wait for @DrTimothyAldenDavis's thoughts on this.

waebbl:
> > do you have an example that illustrates the needs?
> 
> Do you have a list of all projects that use libraries from SuiteSparse?
> 
On Gentoo Linux we have on a quick search eigen (cholmod, umfpack, spqr), scipy (umfpack) and lpsolve (colamd).

> > Could this macro be used to fill in the variables and avoid the `find_package_handle_standard_args`?
> 
> Do you know which variables `find_package_handle_standard_args` sets that we might need to set manually?
> 
From the documentation, I suspect it's the ones we tell the macro to use. What it sets internally, I think we shouldn't care. These variables are usually not exported or cached. The docs say, it's used by `Find*` modules, but we use it from config files. This is where I think, we might go wrong.

> > The cmake build system is quite new to the package, so I wouldn't argue for dropping backward compatibility at this time. IMO this could only be thought about, once the build system has been matured.
> 
> Sorry. I don't get your point. Are you arguing for dropping backward compatibility or against dropping it? Let's wait for @DrTimothyAldenDavis's thoughts on this.

I'm arguing against dropping compatibility right now, but consider dropping it at some later point, once the cmake based build system has been matured.

mmuetzel:
> What it sets internally, I think we shouldn't care. These variables are usually not exported or cached.

There is at least the `*_FOUND` variable that is (or was?) used internally in SuiteSparse. I'm not sure if there won't be more that that function sets.

Edit: Those variables are still used quite extensively even after the changes from this PR. Maybe, I should try and replace them with checks whether the respective targets were imported...

waebbl:
> > What it sets internally, I think we shouldn't care. These variables are usually not exported or cached.
> 
> There is at least the `*_FOUND` variable that is (or was?) used internally in SuiteSparse. I'm not sure if there won't be more that that function sets.

This variable is also set by `find_package(<package> CONFIG)` if the package and all eventually requested COMPONENTS are found.

mmuetzel:
> This variable is also set by `find_package(<package> CONFIG)` if the package and all eventually requested COMPONENTS are found.

I might have misunderstood the following sentence in the documentation:
> The package configuration file may set <PackageName>_FOUND to false to tell find_package that component requirements are not satisfied.

And ISTR that it wasn't set in some tests I did unless I called that function explicitly. But that might have been at some point when the Config.cmake files weren't complete. I'll try to test again.

DrTimothyAldenDavis:
I have some MATLAB scripts that compile MATLAB mexFunctions, and those are configured via config_file in the cmake scripts. Some of those need to know where the libraries are and what they are named. They need these variables.  It's difficult to build a MATLAB mexFunction in a Makefile or CMakeList.txt file so I use a .m file and build them inside MATLAB 

So I do need them. 

SuiteSparse is used in many places but it only recently had any kind of cmake build system.

DrTimothyAldenDavis:
See for example these 2 packages which will soon be added to SuiteSparse:

https://github.com/Aznaveh/SuiteSparse-official/blob/dev2/ParU/Config/paru_matlab_config.m.in
https://github.com/clouren/SPEX/blob/Developmental/SPEX/Config/spex_deps.m.in

Both of these MATLAB scripts are configured by cmake, so that they can find the right libraries and include directories.

waebbl:
From my understanding they could be configured with something like
```
if(TARGET Suitesparse_confg::Suitesparse_config)
  get_target_property(SUITESPARSE_INCDIR Suitesparse_config::Suitesparse_config INTERFACE_INCLUDE_DIRECTORIES)
  get_target_property(SUITESPARSE_LIBDIR Suitesparse_config::Suitesparse_config INTERFACE_LIBRARY_DIRECTORIES)
endif()
```
The include directory is a standard property. If, for the library, the path to the library directory is needed, I think we could set this variable in the `Suitesparse_configConfig.cmake.in` file using something like `set_and_check(INTERFACE_LIBRARY_DIRECTORIES "@SUITESPARSE_LIBRARY_DIR@")`. If the filename to a library is needed, they are part of the generated `*Target*.cmake` files and can be queried through the `IMPORTED_LOCATION{,_RELEASE,_DEBUG,_RELWITHDEBINFO}` properties. We might need to add an `add_library(Suitesparse_config INTERFACE IMPORTED)` call, for this to work.

I didn't meant to say, the required variables should not be set. Just, that I don't think it's a good way to use a module style macro (find_package_handle_standard_args) from a config style cmake file.

mmuetzel:
I had another look at those `*_FOUND` variables. And the only place I "needed" those where they weren't defined without the `find_package_handle_standard_args` macro was inside the the `*Config.cmake` files.
But I copied that idiom over from the `Find*.cmake` modules and, come to think of it, they didn't really make sense here: The `*Config.cmake` files are only created/installed when the package is built/installed (while the former `Find*.cmake` modules were more "standalone"). So, the `else` branch in that expression would never be actually reached.

I retained the checks for the presence of the include directory (where it applies) and the library (as a sanity check). But I agree that the remainder of what that macro did was probably not actually useful.

mmuetzel:
> If, for the library, the path to the library directory is needed, I think we could set this variable in the `Suitesparse_configConfig.cmake.in` file using something like `set_and_check(INTERFACE_LIBRARY_DIRECTORIES "@SUITESPARSE_LIBRARY_DIR@")`.

`SUITESPARSE_LIBRARY_DIR` might be a relative directory with respect to the CMAKE_INSTALL_PREFIX at the time the library was built. I'm not sure if it can be used reliably in dependent projects.

> If the filename to a library is needed, they are part of the generated `*Target*.cmake` files and can be queried through the `IMPORTED_LOCATION{,_RELEASE,_DEBUG,_RELWITHDEBINFO}` properties. We might need to add an `add_library(Suitesparse_config INTERFACE IMPORTED)` call, for this to work.

It was surprisingly complex to get that right for any configuration. But that logic should already be there in the `*Config.cmake.in` files. Would it be better to do that differently?

mmuetzel:
Would it make sense to put all of these imported targets into a `suitesparse::` namespace? Would that help to avoid collisions with target names in downstream projects?

DrTimothyAldenDavis:
That would be a good idea, I think.  I'm not familiar with the module setup in cmake.  So if it makes sense, go ahead, and I'll learn about it.

How would it change how the "find_package" is used by others, to find the packages in SuiteSparse?

mmuetzel:
> How would it change how the "find_package" is used by others, to find the packages in SuiteSparse?

The `find_package` command itself wouldn't need to change. It would still be the same like the one with the `Find*.cmake` modules.
However when referencing the target, the namespace would need to be specified (compared to the current state of this PR). But making the targets available in downstream projects would be new with the changes from this PR anyway. So, this is just a decision whether we would prefer to make those targets available with or without namespace.

mmuetzel:
> > If the filename to a library is needed, they are part of the generated `*Target*.cmake` files and can be queried through the `IMPORTED_LOCATION{,_RELEASE,_DEBUG,_RELWITHDEBINFO}` properties. We might need to add an `add_library(Suitesparse_config INTERFACE IMPORTED)` call, for this to work.
> 
> It was surprisingly complex to get that right for any configuration. But that logic should already be there in the `*Config.cmake.in` files. Would it be better to do that differently?

It turned out that logic wasn't right for cases where SuiteSparse and dependent projects were configured for different `CMAKE_BUILD_TYPE`s. E.g., if SuiteSparse was configured for let's say `Release` but dependent projects were configured for `Debug`, the library detection would fail.
That issue should be fixed with the most recent update.

mmuetzel:
There might be multiple include directories and/or libraries in the variables generated by the *Config.cmake files. And some of them might still be generator expressions.
The recent set of changes accounts for that.

Additionally, it now also includes changes for GraphBLAS.

mmuetzel:
Rebased on a current head of the `dev2` branch.

mmuetzel:
I don't know why cmake was able to find the installed OpenBLAS for the example without any additional hints just a few days ago for #290. But in the latest build attempt here, it didn't detect it...

mmuetzel:
Rebased on a current head of `dev2`.
Let's see how many mistakes I made while trying to resolve the conflicts. 😉

K20shores:
@mmuetzel @DrTimothyAldenDavis, I also have an interest in these PRs related to cmake. It would make consuming SuiteSparse easier for some of the projects I work on. Is there any way that I can help?

mmuetzel:
This PR is waiting for the changes in #305 and #306.

Progress has stalled since a couple of months. I presume because @DrTimothyAldenDavis is otherwise occupied.

DrTimothyAldenDavis:
Yes -- I've been tied up with some tight deadlines but I haven't forgotten about the updates to the cmake build system for SuiteSparse.  I should be able to get back to it sometime later this week.

mmuetzel:
As @chengts95 pointed out in #309, there might be an issue if the same or a newer version of SuiteSparse is already installed on the system where SuiteSparse is built.
I updated the rules to always prefer the targets that were built in a common source tree (i.e., using the top-level `Makefile`).

I hope I didn't miss some of the possibly affected instances of `find_package`.

mmuetzel:
I rebased on a current head of `dev2`. Many conflicts needed to be resolved. Let's see what the CI is saying...

@DrTimothyAldenDavis: Thank you very much for merging the other PRs. 🚀

mmuetzel:
This PR should be pretty close to being ready for review finally.

Some remaining points I can think of:
1. The build rules for the example could be simplified if #336 is applied.
2. All targets could be put into a `suitesparse::` namespace (see https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/267#issuecomment-1377721030 and https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/267#issuecomment-1382314115).
3. Test if this change is noticeable in existing downstream projects using Cmake. (It shouldn't be noticeable if everything is working as expected.)

Points 1 and 2 could be made in this PR or in follow-up changes. I slightly tend towards a follow-up PR for point 2 because this PR is already pretty large as it is.
Point 1 could be done at any time. Integrating it into this PR might be slightly nicer imho. 
It might be better to make the change for point 2 before a version with the changes from here is released. (I.e., it might be best to not apply this PR if a release is imminent.) Otherwise, there would be versions for which the import targets are not within a namespace and newer ones where they are in a namespace. That might make it awkward to use in downstream projects.

It might be slightly easier to test if everything works as expected when this PR is already applied (point 3).

@DrTimothyAldenDavis: What are your plans or approximate timeline for the next minor or major releases?

mmuetzel:
Ah. And potentially a rather big point:
4. Where should the available targets be documented? (Or is the example good enough for that?)

mmuetzel:
Rebased on a current head of `dev2` including the changes from #336. Let's wait for the CI.

mmuetzel:
The CI built and ran the example on all available platforms. Other than that, there are no changes to before the rebase.

Marking this as ready for review. But like written before and in another PR, it shouldn't be merged before v7.1.0.

DrTimothyAldenDavis:
Other than the merge conflict, I'm ready to give this a try in dev2.  There are merge conflicts from the NSTATIC changes from @kiwifb that were added to dev2 ahead of this update.  I'd like to keep those changes to the NSTATIC handling in these cmake scripts, of course.

 @mmuetzel : shall I merge it in and fix the conflicts?  Or would you like to fix those conflicts?  Is it ready for me to merge and give a try?

mmuetzel:
I can try to rebase this in a couple of days of you don't beat me to it.

DrTimothyAldenDavis:
I can take a stab at it.  It looks straightforward ... mostly your edits would take precedence, but building the static cuda libraries would be placed inside "if ( NSTATIC ) ...".  If I break something, there's always the original PRs I can refer back to.

DrTimothyAldenDavis:
OK ... I think that's the resolution.  I left the indenting as-is, to keep the edits short.  I'll let the CI run the checks, and when it passes, I'll merge in this PR.  Thanks again for the help!

DrTimothyAldenDavis:
Looks good so far.  I did find one glitch in the SPQR CMakeLists.txt file that I fixed here:    https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/bcf8ec4157fb73c679140c176b0b3ca7acadd52e .

That fix looks pretty straightforward.  Prior to this fix, I got a warning on line 98, about CCOLAMD not being found.  It was then found by the subsequent logic, but the first find_package statement was causing the failure, and it didn't match the pattern of the rest of the packages.

PR #275, opened 2023-01-11
Remove Makefile target `both`

mmuetzel:
Like discussed in https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/267#issuecomment-1377758148, the Makefile target `both` is not strictly necessary. Removing it would simplify subsequent changes a lot.

With the proposed changes, it will still be possible to install to the local and global destination (without recompiling). For this, the following sequence of commands will be necessary:
```
make global
make install
make local
make install
```

The intermediate `make local` step is only needed for reconfiguring for the different prefix. That command shouldn't trigger any recompilation and should thus be fast.
Installing to the local prefix first and then to the global prefix would also be possible by inverting the respective commands in the above sequence.

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -132,37 +132,17 @@ endif ( )
 # AMD installation location
 #-------------------------------------------------------------------------------
 
-if ( GLOBAL_INSTALL )
-    # install in /usr/local/lib and /usr/local/include.
-    # requires "sudo make install"
-    message ( STATUS "Installation will be system-wide (requires 'sudo make install')" )
-    install ( TARGETS amd
-        LIBRARY       DESTINATION ${CMAKE_INSTALL_LIBDIR}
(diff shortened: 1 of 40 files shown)


DrTimothyAldenDavis:
Looks great - thanks.  I can add this to v7.0.0.  The CMake scripts are simpler with this change, and I don't really need "make both".  Doing the 4 steps "make ; make install ; make local ; make install" is fine, if I want to use my top-level Makefile.

mmuetzel:
> I can add this to v7.0.0.

I wasn't expecting this part to already ride the train for v7.0.0.
I noticed a couple of subprojects that I forgot to modify. I'll post a PR for those shortly.
Also GraphBLAS isn't completely adapted yet to the new local/global selection. Should I adapt it here? Or in its own repository?

mmuetzel:
See #277.

PR #276, opened 2023-01-11
Use function pointers instead of void* for get and set functions.

mmuetzel:
I built and installed the libraries from the `7.0.0.beta1` tag. 
I tried to rebuild Octave with them installed. But (unsurprisingly) it needed some changes to account for the new API with get and set functions instead of the global structure.
Casting forth and back between function pointers and `void *` is pretty awkward in C++ if you try to avoid old-style casts. (`reinterpret_cast` fails on `nullptr` and `static_cast` fails on function pointers.)
Would it be possible to avoid the cast to `void *` in the API? That would also improve type safety. When a user inadvertently passes the pointer to a function with a non-matching signature, the compiler would already complain (instead of a crash on runtime).

The proposed changes use function pointers as the arguments and return values of the newly added get and set functions.

### DIFF
--- a/CHOLMOD/Partition/cholmod_metis.c
+++ b/CHOLMOD/Partition/cholmod_metis.c
@@ -98,18 +98,18 @@
         save_calloc_func  = SuiteSparse_config_calloc_func_get ( ) ;        \
         save_realloc_func = SuiteSparse_config_realloc_func_get ( ) ;       \
         save_free_func    = SuiteSparse_config_free_func_get ( ) ;          \
-        SuiteSparse_config_malloc_func_set ((void *) malloc) ;              \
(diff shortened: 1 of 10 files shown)


mmuetzel:
(I should probably add that I was able to rebuild (a modified) Octave with these changes. And no new tests are failing.)

mmuetzel:
It would be nice if this could be added before SuiteSparse 7.0.0 will be released.

DrTimothyAldenDavis:
Thanks -- that's a good idea.  I find the C types of parameters like these pretty inscrutable so I just used (void *) for simplicity, but these function signatures are better.  I'll add this to v7.0.0.

PR #277, opened 2023-01-12
Adjust install rules of sub-projects missing from #275.

mmuetzel:
I missed a few files that should no longer depend on `GLOBAL_INSTALL`.

There are still a few places in GraphBLAS where that variable is used. But it has its own logic to instantiate that cmake variable.
Should those changes be made in this repository? Or in the GraphBLAS repository?


DrTimothyAldenDavis:
No problem.

I thought this change was needed for v7.0.0.  It looks like the right thing to do, so if it's working we can keep it for v7.0.0 and get the rest changed.  If it's too much to do now, we could back out this change.  But it doesn't look like it would be too hard to get into v7.0.0.

For GraphBLAS: go ahead and change it here, not in the GraphBLAS repo.  Then I can copy over the changes from here to there, once v7.0.0 is released.

GraphBLAS has a copy of SuiteSparsePolicy.cmake in its cmake_modules folder.  That should be an exact copy of the one in SuiteSparse_config.

mmuetzel:
You're right. It wasn't too many places that needed changing. (I hope I didn't miss more.)

The cmake modules already differed a bit. I didn't bother to pick the differences but just copied the version from inside SuiteSparse_config over to GraphBLAS. I hope that is ok.

DrTimothyAldenDavis:
Yes, that looks correct.  I'll test it all out.

PR #278, opened 2023-01-15
Mongoose: Support building no static library.

mmuetzel:
Rename the targets to align with other SuiteSparse libraries.
This part of the changes is entirely internal currently. But the names of the targets will be facing the public if #267 is applied.

### DIFF
--- a/GraphBLAS/cmake_modules/SuiteSparsePolicy.cmake
+++ b/GraphBLAS/cmake_modules/SuiteSparsePolicy.cmake
@@ -29,9 +29,7 @@
 #   NSTATIC:            if true, static libraries are not built.
 #                       Default: false, except for GraphBLAS, which
 #                       takes a long time to compile so the default for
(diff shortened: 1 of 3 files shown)


mmuetzel:
@DrTimothyAldenDavis: Would you prefer for me to split this into two different PRs? One about the target name change and a separate one about building no static library?

DrTimothyAldenDavis:
Thanks for the update.  Yes, the Mongoose cmake build system is a little different from the others.  It was written first, a while back, by Scott Kolodziej, before I knew cmake.  I have revised it a bit, but not much.  So this revision is a good idea.

A single PR to dev2 is fine -- I just haven't gotten to trying this out yet.  I wanted to finish polishing up and releasing a stable SuiteSparse v7.0.1 first.  I should be able to merge this PR in in the next few days.

mmuetzel:
@DrTimothyAldenDavis: Did you run into issues with this PR?

DrTimothyAldenDavis:
No ... I'm just behind on many other deadlines.

PR #284, opened 2023-01-21
Example: update cmake modules and LDL version.

mmuetzel:
The example currently doesn't compile on MinGW (at least with the packages currently provided by the MSYS2 project).

This PR copies the respective cmake modules from SPEX and updates the version number of LDL in the example.


mmuetzel:
Related question: Would it make sense to install the libraries and build the example as part of the CI?

DrTimothyAldenDavis:
Thanks for the fix.  Yes, it would be a good idea.  The Example package is a good test for all of the Find*.cmake or *Config.cmake scripts.

DrTimothyAldenDavis:
I found a minor glitch in GraphBLAS (now v7.4.3) and will need to update SuiteSparse to v7.0.1 with this minor fix.  I can include this fix to the Example package in v7.0.1 as well.

PR #287, opened 2023-01-22
Build example as part of the CI

mmuetzel:
This PR adds steps to build (and execute) the example as part of the CI.
This helps to check if the Find*.cmake modules are doing the "Right Thing™".

Adding this revealed a few (potential) issues:
- The `ubuntu (clang)` runner was building without OpenMP. This only affected the CI and should be fixed with the changes here.
- Using the Find*.cmake modules with MSVC didn't fall back to the shared library if the static library wasn't built. That is the default for the GraphBLAS library. This currently also affects downstream users with MSVC. This should be fixed by the second commit in this PR.
- It would be nice if some of the preprocessor flags would be public. E.g., downstream users of CXSparse would likely be interested in whether the library was built with `-DNCOMPLEX` in the preprocessor/compiler flags. (It might also be nice if that preprocessor flag would be renamed to something more specific like `CXSPARSE_NCOMPLEX` to avoid symbol collisions in downstream projects.) There are no changes related to this in this PR. It might be easier to tackle this once/if downstream could import targets (see #267).

Edit: For an alternative approach to avoid the issue described in the last point see #288.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -33,7 +33,7 @@ jobs:
             cc: "gcc"
             cxx: "g++"
           - compiler: clang
-            compiler-pkgs: "clang"
+            compiler-pkgs: "clang libomp-dev"
             cc: "clang"
             cxx: "clang++"
           # Clang seems to generally require less cache size (smaller object files?).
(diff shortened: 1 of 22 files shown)


DrTimothyAldenDavis:
Can I safely move this to the dev2 branch?

mmuetzel:
> Can I safely move this to the dev2 branch?

Done.

PR #288, opened 2023-01-22
CXSparse: Match API in public header to compiled library.

mmuetzel:
Currently, a downstream user would need to know whether the CXSparse library was built with or without `NCOMPLEX`. The user would need to compile their dependent project with that preprocessor value defined (or undefined) accordingly. That might be difficult to determine in general, e.g., if the library was downloaded from a binary distribution repository (i.e., most Linux distributions).

The proposed changes hard-code the value that was used during compilation in the public header. So, a downstream user would no longer need to worry about the status of that flag during compilation.

Edit: By removing that preprocessor macro from the public header, the risk of a symbol collision in downstream projects would also be avoided.

### DIFF
--- a/CXSparse/CMakeLists.txt
+++ b/CXSparse/CMakeLists.txt
@@ -46,8 +46,10 @@ endif ( )
 if ( NCOMPLEX )
     message ( STATUS "complex data type: disabled" )
     add_compile_definitions ( NCOMPLEX )
+    set ( CXSPARSE_USE_COMPLEX 0 )
 else ( )
     message ( STATUS "complex data type: enabled" )
+    set ( CXSPARSE_USE_COMPLEX 1 )
 endif ( )
 
 #-------------------------------------------------------------------------------
--- a/CXSparse/Config/cs.h.in
+++ b/CXSparse/Config/cs.h.in
@@ -27,13 +27,13 @@
 #define _CXS_H
 
 #ifdef __cplusplus
-#ifndef NCOMPLEX
+#if @CXSPARSE_USE_COMPLEX@
 #include <complex>
 typedef std::complex<double> cs_complex_t ;
 #endif
 extern "C" {
 #else
-#ifndef NCOMPLEX
+#if @CXSPARSE_USE_COMPLEX@
 #include <complex.h>
 #define cs_complex_t double complex
 #endif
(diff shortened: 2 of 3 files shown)


DrTimothyAldenDavis:
That's a good idea.  I'd like to have this apply to the dev2 branch, though.  I'd like to make a small revision and I prefer to do all that in dev2, and then push to dev when I think it's ready for the world to see as the default in-progress branch (with all features working, in place, but not yet released).  Can I safely retarget this PR to dev2?

mmuetzel:
> Can I safely retarget this PR to dev2?

Oops. Forgot to change to dev2 from the default branch (which is automatically selected when creating a PR). I've probably also forgotten to do the same for other PRs. Sorry for that.

DrTimothyAldenDavis:
No problem.  I will likely add a safe #define that is visible to the user, so they can query at compile time if the features are available, such as 

`#ifdef CS_HAS_COMPLEX`
...

or `#ifdef CXSPARSE_HAS_COMPLEX`

or something appropriate.

mmuetzel:
> No problem. I will likely add a safe #define that is visible to the user, so they can query at compile time if the features are available, such as
> 
> `#ifdef CS_HAS_COMPLEX` ...
> 
> or `#ifdef CXSPARSE_HAS_COMPLEX`
> 
> or something appropriate.

Good idea. But I'm not sure if `HAS` is the right word. At least in GNU projects, `HAVE_` in preprocessor macros means that something is there *in principle* (a library or header is installed or an OS supports a feature). But it might not actually be used. `USE_` indicates that a user chose to use or not use something. That could be out of necessity or a "free decision".
IIUC, this case would better fit to `USE_`.

DrTimothyAldenDavis:
Sounds reasonable.  I could make `CXSPARSE_USE_COMPLEX` a term that is #define'd if the complex functions are available,and not #define'd otherwise.

DrTimothyAldenDavis:
Or I could always define it but do either

`#define CXSPARSE_USE_COMPLEX 1`

if available, or

`#define CXSPARSE_USE_COMPLEX 0`

if not.

DrTimothyAldenDavis:
By adding:

`#define CXSPARSE_USE_COMPLEX @CXSPARSE_USE_COMPLEX@`

in cs.h.in, for example.

mmuetzel:
> By adding:
> 
> `#define CXSPARSE_USE_COMPLEX @CXSPARSE_USE_COMPLEX@`
> 
> in cs.h.in, for example.

That sounds reasonable. I like it. 👍

PR #293, opened 2023-02-07
Add missing bracket to close extern C.

samuelburbulla:
No pull request can be accepted unless you first sign the Contributor License Agreement [ CONTRIBUTOR-LICENSE.txt ].  Print it as a PDF and email me a signed PDF (digital signature OK).  Submit all PRs to the dev2 branch only.


DrTimothyAldenDavis:
Oops.  Thanks for the catch.  The cmake build script creates this file from LDL/Config/ldl.h.in, though, and will overwrite this change.  No problem though -- I can copy the change over to that file.

PR #294, opened 2023-02-08
Add C++ example

mmuetzel:
Add an example that includes all headers in code that is compiled with a C++ compiler.
This will hopefully help to reveal issues like the one addressed by #293.

The example itself is almost identical to the example in C. There are only minor changes to use C++ idioms here and there.

Adding this also revealed that there are a couple of headers that didn't mark their function declarations with C linkage (GraphBLAS and SPEX).

I also included a fix similar to #293 to have the CI still passing with the new example.

### DIFF
--- a/Example/CMakeLists.txt
+++ b/Example/CMakeLists.txt
@@ -51,7 +51,7 @@ set ( CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
 
 project ( my
     VERSION "${MY_VERSION_MAJOR}.${MY_VERSION_MINOR}.${MY_VERSION_PATCH}"
-    LANGUAGES C )
+    LANGUAGES C CXX )
 
 #-------------------------------------------------------------------------------
 # find library dependencies
@@ -101,7 +101,7 @@ configure_file ( "Config/my.h.in"
 include_directories ( Source Include )
 
 #-------------------------------------------------------------------------------
-# my dynamic library properties
+# my dynamic C library properties
 #-------------------------------------------------------------------------------
 
 file ( GLOB MY_SOURCES "Source/*.c" )
@@ -115,7 +115,7 @@ set_target_properties ( my PROPERTIES
     WINDOWS_EXPORT_ALL_SYMBOLS ON )
 
 #-------------------------------------------------------------------------------
-# my static library properties
+# my static C library properties
 #-------------------------------------------------------------------------------
 
 add_library ( my_static STATIC ${MY_SOURCES} )
@@ -132,6 +132,37 @@ if ( MSVC )
 endif ( )
 
 #-------------------------------------------------------------------------------
+# my dynamic C++ library properties
+#-------------------------------------------------------------------------------
+
+file ( GLOB MY_SOURCES "Source/*.cc" )
+add_library ( my_cxx SHARED ${MY_SOURCES} )
+set_target_properties ( my_cxx PROPERTIES
+    VERSION ${MY_VERSION_MAJOR}.${MY_VERSION_MINOR}.${MY_VERSION_PATCH}
+    CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED ON
+    SOVERSION ${MY_VERSION_MAJOR}
+    PUBLIC_HEADER "Include/my.h"
+    WINDOWS_EXPORT_ALL_SYMBOLS ON )
+
+#-------------------------------------------------------------------------------
+# my static C++ library properties
+#-------------------------------------------------------------------------------
+
+add_library ( my_cxx_static STATIC ${MY_SOURCES} )
+set_target_properties ( my_cxx_static PROPERTIES
+    VERSION ${MY_VERSION_MAJOR}.${MY_VERSION_MINOR}.${MY_VERSION_PATCH}
+    CXX_STANDARD 11
+    CXX_STANDARD_REQUIRED ON
+    OUTPUT_NAME my_cxx
+    SOVERSION ${MY_VERSION_MAJOR} )
+
+if ( MSVC )
+    set_target_properties ( my_cxx_static PROPERTIES
+        OUTPUT_NAME my_cxx_static )
+endif ( )
+
+#-------------------------------------------------------------------------------
 # add the library dependencies
 #-------------------------------------------------------------------------------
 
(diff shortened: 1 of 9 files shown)


mmuetzel:
It looks like Clang and MSVC are stricter than GCC when it comes to passing `const char *` variables as `char *` arguments.
Looking at the definition of RBio's `RBwrite` function, the first three arguments should in fact be `const char *` (but they are `char *` currently).
For the time being manually cast away the `const` qualifier when calling that function in the example. The "real" fix would be changing the prototype of that function. But that might constitute an ABI change (and is probably out of scope of this PR).

mmuetzel:
Clang emits warnings like the following for the C++ example:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/4132008103/jobs/7140206808#step:12:264
```
  In file included from D:/a/SuiteSparse/SuiteSparse/Example/Source/my_cpp.cc:7:
  In file included from D:/a/SuiteSparse/SuiteSparse/Example/Include/my.h:20:
  D:/a/SuiteSparse/SuiteSparse/CXSparse/include/cs.h:451:14: warning: 'cs_ci_house' has C-linkage specified, but returns user-defined type 'cs_complex_t' (aka 'complex<double>') which is incompatible with C [-Wreturn-type-c-linkage]
  cs_complex_t cs_ci_house (cs_complex_t *x, double *beta, int32_t n) ;
               ^
```

It is probably right to warn about this. But I haven't looked into it.
@DrTimothyAldenDavis: Do you know why `cs_complex_t` is a different type depending on whether the header is included in C or C++? Which is the "correct" type that is used by the exported functions?

Edit: Similar warnings from MSVC:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/4132008103/jobs/7140206906#step:18:314
```
    my_cpp.cc
  D:\a\SuiteSparse\SuiteSparse\include\GraphBLAS.h(12070,60): warning C4190: 'GxB_Iterator_get_FC32' has C-linkage specified, but returns UDT 'std::complex<float>' which is incompatible with C [D:\a\SuiteSparse\SuiteSparse\Example\my_cpp.vcxproj]
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.34.31933\include\complex(1007,22): message : see declaration of 'std::complex<float>' [D:\a\SuiteSparse\SuiteSparse\Example\my_cpp.vcxproj]
```

mmuetzel:
My initial naming (`my_cpp`) of the C++ example was a bit unfortunate (`CPP` is mostly used to refer to the C pre-processor). I renamed it to `my_cxx` which might be a bit more "in line" with the usage of `CXX` for the C++ compiler.

mmuetzel:
My initial naming (`my_cpp`) of the C++ example was a bit unfortunate (`CPP` is mostly used to refer to the C pre-processor). I renamed it to `my_cxx` which might be a bit more "in line" with the usage of `CXX` for the C++ compiler.

This PR should be ready for review now.

DrTimothyAldenDavis:
I'll have to take a closer look later.

For C vs C++:   the type has to match the compiler's native complex type, which is 'double complex' for C and std::complex<double> for C++.   It makes sense to use the native type for each particular language.

Perhaps in that case, I shouldn't use "extern "C" { ... }" in cs.h, and let C++ compile all of CXSparse.  Either that, or I still compile CXSparse in C but work out some other solution to mute this warning.

mmuetzel:
> For C vs C++: the type has to match the compiler's native complex type, which is 'double complex' for C and std::complex for C++. It makes sense to use the native type for each particular language.

I'm not sure if I can follow that statement. But maybe it is better to defer that to until after this PR is (potentially) merged. That might make it easier to explore potential changes and have the CI check them automatically with a bunch of different compilers/platforms.

For now, I rebased on a current head of the `dev2` branch.

mmuetzel:
The last commit in this PR ("Example: Cast away `const` qualifier for arguments of `RBwrite`.") can probably be omitted (or reverted) in case #297 should be accepted.

mmuetzel:
Coming to this again because I stumbled upon a good reference:
> For C vs C++: the type has to match the compiler's native complex type, which is 'double complex' for C and std::complex for C++. It makes sense to use the native type for each particular language.

For a good reason why it might not be a good idea to use `std::complex<double>` when declaring the functions in the header but actually using `complex double` when the functions are compiled, see e.g.:
https://reviews.llvm.org/D83397#2169102 (point 3 there)

Citing here for easier reference:
> `_Complex` and `std::complex` are indeed layout compatible, but they are not guaranteed to be compatible in call interface. `std::complex<>` is an aggregate type that is opaque to the compiler, `_Complex` is a fundamental type for which some ABIs have different interface requirements.

Applying that to the case here: It is ok for C++ code to call a function that has `_Complex` arguments with variables of type `std::complex<>`. The compiler will know how to pass them. But that is only true if the compiler actually knows that the function expects `_Complex` type arguments.
Hiding the actual calling interface of the function in the header could lead to issues on some platforms (e.g., x86 32-bit) ~~with some compilers~~.

mmuetzel:
I opened PR #312 for the issue discussed in the previous comment.

DrTimothyAldenDavis:
I'll have to see how this affects GraphBLAS as well, and the CUDA kernels in particular (which are in C++).  We don't have the complex data types working in CUDA yet, however.

Those CUDA kernels will be operating on complex data types on the GPU, and they will need to use std::complex because they'll have to use all kinds of C++ functions (std::sin, cos, atan, etc, on complex data types).

When those results are sent back to the CPU side, the GraphBLAS C code will need to read those results, and they will be in "double complex" or whatever.

DrTimothyAldenDavis:
See in particular these files:

https://github.com/DrTimothyAldenDavis/GraphBLAS/blob/master/Source/Shared/GB_complex.h
https://github.com/DrTimothyAldenDavis/GraphBLAS/blob/master/Source/Shared/GxB_complex.h

If we can get everything to work with C++ seeing the data types as C complex types, as in:

`extern "C" { typedef GxB_FC32_t float complex ; }
`
then this change makes sense.  It would mean that everyone (C, C++, and Microsoft C) would agree that the underlying data type is a C-style float complex and double complex.

I'm just unsure how it will work, or if these changes will break the GraphBLAS CUDA kernels.

I bring up GraphBLAS here because whatever I do for CXSparse, I should do the same for all other C/C++ mixtures in the rest of SuiteSparse (including at least CXSparse, SPQR, and GraphBLAS).

DrTimothyAldenDavis:
Another thing:  I don't often return complex scalars from any of my user-callable functions, as return values.    The only exception that I've found so far is the cs_*_house function in CXSparse, and if necessary I could change that.  I might have other cases.

If using the C complex type in the prototypes is more robust when calling the C functions from C++, then it's worth getting it to work.  I'll need to bump the major SO numbers for many of my codes, because it changes the user-visible API.  I'll want to make sure it works in GraphBLAS in the CUDA kernels as well, but I'm guessing I can fix that later (it doesn't work now anyway) since none of the CUDA kernels are user-visible anyway.  All user-visible access to GraphBLAS is through C.

mmuetzel:
I don't understand how adding an example in C++ would have an adverse effect on GraphBLAS or other SuiteSparse libraries.
Which specific change do you think could be dangerous?

DrTimothyAldenDavis:
Adding the C++ example isn't the issue.  The issue is that if I change the complex type in CXSparse, I should be consistent and do the same for GraphBLAS, but that might cause issues with the CUDA kernels.

But I think I should go ahead with your change to CXSparse (#312).   It looks like the safer, more stable, thing to do.  I'll work on GraphBLAS to see how it plays out there.

mmuetzel:
Thank you for pushing #312.
I rebased on top of the current head of `dev2` to check if `clang++` still warns. (It should no longer.)

mmuetzel:
Thank you for pushing #318.
Rebased this PR on top of those changes. Let's see how it'll fare now.

Edit: Builds without the warning now. See also https://github.com/DrTimothyAldenDavis/GraphBLAS/pull/220 for the warning from the GraphBLAS code.

DrTimothyAldenDavis:
Thanks for the updates.  I'm finally been able to free up my schedule so I can focus on the SuiteSparse meta-package again.

PR #295, opened 2023-02-09
Set the required C/C++ standard in cmake rules.

mmuetzel:
Currently, the Boolean target properties `C_STANDARD_REQUIRED` and `CXX_STANDARD_REQUIRED` are set to a non-0 value without specifying which standard is required.
Assuming the value that is used to set the Boolean property is meant to specify the standard, this PR sets the respective value to the properties `C_STANDARD` or `CXX_STANDARD` and uses the ("more default") value `ON` for the Boolean properties.


DrTimothyAldenDavis:
Yes, this is a good idea ... I misunderstood the meaning of those variables.

PR #297, opened 2023-02-16
RBio: Use pointers to constant PODs in interface where applicable.

mmuetzel:
Some of the pointer arguments of the RBio functions aren't changed by the respective functions. But since they aren't `const type*` arguments, it is a compiler error to pass arguments that are constant.

This PR adds the `const` qualifier to those pointer arguments. Among the advantages are that user code could use (safer) `const` arguments, as a consequence the API would be safer, and the compiler might be able to produce more efficient code in applications using these functions.

Adding the `const` qualifier doesn't change the API afaict. Code using these functions doesn't need to be changed. 
I'm not sure if those changes affect the ABI on any platform though. It might be necessary to recompile user code (on some platforms?) if those changes should be applied. (But tbh, I'm not sure about that.)

Edit: Function pointers in user code might need to be adapted. So, this change probably qualifies as an API change (contrary to what I previously wrote).

### DIFF
--- a/RBio/Config/RBio.h.in
+++ b/RBio/Config/RBio.h.in
@@ -114,12 +114,12 @@ extern "C" {
 int RBkind_i        /* 0: OK, < 0: error, > 0: warning */
 (
     /* input */
-    int32_t nrow,   /* A is nrow-by-ncol */
+    int32_t nrow,       /* A is nrow-by-ncol */
     int32_t ncol,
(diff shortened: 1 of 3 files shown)


DrTimothyAldenDavis:
Thanks for the updates.  I think it would be safest to update the major SO number for RBio, so this would become v4.0.0.  I'll merge this into dev2 and try it out.

PR #305, opened 2023-02-26
Correctly save and restore CMAKE_FIND_LIBRARY_SUFFIXES for GMP and MPFR modules

mmuetzel:
Currently, `CMAKE_FIND_LIBRARY_SUFFIXES` is (re-)set to the value of `save` in the modules `FindGMP.cmake` and `FindMPFR.cmake`. But that variable might not have been set depending on where those functions are being called. This means that `CMAKE_FIND_LIBRARY_SUFFIXES` is effectively set to the empty string `""`.
This can cause issues for subsequent calls to `find_library`.

The intention was probably to change `CMAKE_FIND_LIBRARY_SUFFIXES` for the duration of the call to `find_library` in those modules and reset it to its original value afterwards.

This PR does that and also changes the name of the variable `save` to something that is less likely to collide with a variable from the context of the caller.

This would probably fix the failing CI in #267.

### DIFF
--- a/Example/cmake_modules/FindGMP.cmake
+++ b/Example/cmake_modules/FindGMP.cmake
@@ -44,6 +44,7 @@ find_library ( GMP_LIBRARY
 
 # static gmp library
 if ( NOT MSVC )
+    set ( save_CMAKE_FIND_LIBRARY_SUFFIXES ${CMAKE_FIND_LIBRARY_SUFFIXES} )
     set ( CMAKE_FIND_LIBRARY_SUFFIXES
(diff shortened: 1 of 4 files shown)


mmuetzel:
This is a pretty localized change that might be easy to review:
@DrTimothyAldenDavis: If you come around to it could you please check if this can be merged? That way I could rebase #267 on top of this.

DrTimothyAldenDavis:
Got it -- sorry for the delay, and thanks again for all your help.

PR #306, opened 2023-02-26
Don't propagate linked libraries in target for shared libraries

mmuetzel:
Currently, most of the `target_link_libraries` commands add libraries to the respective target with the `PUBLIC` keyword. But in general, dependencies of shared libraries don't need to be linked in explicitly. Doing that could lead to overlinking the resulting binaries.
That isn't an issue currently (despite for some demo executables). But it might become one in case #267 will be merged.

This PR switches (most of) the `target_link_libraries` commands on targets for shared libraries to `PRIVATE` (first commit). Instead of implicitly linking those libraries to all dependent targets, it explicitly adds those dependencies for the targets that actually need them (remaining commits).

Currently, this change is only internal without any effect on projects linking any of the SuiteSparse libraries. But it will make a difference if #267 is merged (and dependent projects might start linking to the cmake targets).


mmuetzel:
@DrTimothyAldenDavis: I understand that your time is stretched. Since there are several open PRs, I hope it will help to give them different priority.
If you have some time to spare for SuiteSparse, I'd appreciate if you could review this PR first. I'd like to rebase #267 on top of this PR if it is merged.

mmuetzel:
Rebased on a current head of the `dev2` branch.

DrTimothyAldenDavis (approved):

DrTimothyAldenDavis:
Looks great -- thanks.

PR #312, opened 2023-04-22
CXSparse: Use correct type in declaration of functions in C++ code.

mmuetzel:
Use the same type in function declarations that was used in the function definition on compile time.

See also the following comment for why this is the "right thing" here:
https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/294#issuecomment-1518140344


mmuetzel:
Thank you for merging this.

PR #316, opened 2023-05-11
Attempt to Fix 32-bit 

rayegun:
This is a messy first pass at fixing the 32-bit issues. It still needs a bit of tweaking and notably requires C++11 in its current form. If that is alright then I think it "just" needs some correctness testing that I can't seem to run on my Mac.

It's quite a large diff, and there are better ways to do this, but it seemed to me the least amount of upheaval, especially in the public API.

I also need to go over it with a fine toothed comb, since 98% of it was just find and replace.

Oh! There's CI that's awesome


mmuetzel:
Thank you for taking this on.

I haven't tested the changes yet. Just perused the diff.
It looks like it would break backward compatibility by replacing the symbols that currently correspond to classes by class templates.

I haven't tested that. But would the following pattern work to keep backward compatibility?
In the header:
```
// define class template
template <typename Int = size_t> struct spqr_gpu_impl
{
// ...
};

// import template class from library (or other compilation unit)
extern template struct spqr_gpu_impl<int32_t>;
extern template struct spqr_gpu_impl<size_t>;

// typedef with (existing) name
typedef spqr_gpu_impl<int32_t> spqr_gpu;
typedef spqr_gpu_impl<size_t> spqr_gpu_l;
```

And in the C++ file:
```
// templated (member) function definitions
// ...

// specializations if needed
// ...

// explicit instantiations of template classes
template struct spqr_gpu_impl<int32_t>;
template struct spqr_gpu_impl<size_t>;
```

(Btw, I used `size_t` instead of `int64_t` because I think that would be the "natural" choice on 32-bit and 64-bit platforms. That is different from the other libraries in SuiteSparse though.)

SPQR/Source/spqr_cholmod_wrappers.cpp
mmuetzel, line 1: Would it make sense to include those specializations in the header to allow the compiler to inline them?

Wimmerer:
Hey,

C++ is far from my primary language so apologies in advance, the subtleties sometimes escape me.

Does the default template not make it compatible? 

I chose int32_t and int64_t so that on 64-bit platforms both are available. On 32-bit platforms it is perhaps unwise to maintain the 64-bit symbols but I think it's more reasonable than removing them.

rayegun:
Hey,

C++ is far from my primary language so apologies in advance, the subtleties sometimes escape me.

Does the default template not make it compatible? Perhaps when it's a member of another struct? 

I chose int32_t and int64_t so that on 64-bit platforms both are available. On 32-bit platforms it is perhaps unwise to maintain the 64-bit symbols but I think it's more reasonable than removing them. 

Oh as the default. Sure that might be fine let me look.

mmuetzel:
> Does the default template not make it compatible? Perhaps when it's a member of another struct?

Existing code might be using something like `spqr_gpu a;`. If I'm not mistaken that's only legal for templates since C++17. Before that, you'd need to use something like `spqr_gpu<> a;` to instantiate with default template parameters.

rayegun:
Oh I see about the parameter now, okay let me look at trying to fix that.

mmuetzel:
> I chose int32_t and int64_t so that on 64-bit platforms both are available. On 32-bit platforms it is perhaps unwise to maintain the 64-bit symbols but I think it's more reasonable than removing them.

I agree with that. That's why I proposed `int32_t` and `size_t`.
With these two, 64-bit platforms would have functions for 32-bit and 64-bit integers.
32-bit platforms would have the same set of symbols as 64-bit platforms. But both would use (actually sensibly usable) 32-bit integers. Downstream projects would not need to maintain a separate code path for 32-bit platforms. They could either chose the 32-bit interface for 32-bit and 64-bit targets, or the interface with the "largest meaningful" integer type for indexing for any target.

rayegun:
@DrTimothyAldenDavis is C++11 compat minimum acceptable? I think that's what the CMake build system says but I want to check

rayegun:
@DrTimothyAldenDavis also, should these integer inputs always be signed, or is `size_t` okay? If we need signing is `using signed_size_t = std::make_signed_t<std::size_t>;` acceptable?

DrTimothyAldenDavis:
I can't use size_t as an integer.  Unsigned arithmetic is a problem, and all my codes assume the integer used for indices is signed, not unsigned.

A signed size_t is a problem too.  It doesn't exist in C.

I need to stick with int32_t and int64_t as the two integers.  I want to be able to use GraphBLAS and my solvers together, for example, and on a 32-bit platform that means I would need to use int64_t, not size_t.

It's also very confusing to have an integer like size_t (or a signed variant) that differs from platform to platform.  This confusion will likely lead to things breaking across SuiteSparse.

In my other solvers, and in GraphBLAS, I never use a signed size_t of any kind.  Just int32_t and int64_t.  It's too disruptive to change to another strategy just for SPQR.

DrTimothyAldenDavis:
> 



> @DrTimothyAldenDavis is C++11 compat minimum acceptable? I think that's what the CMake build system says but I want to check

I think C++11 should be OK.

rayegun:
Okay, I'll stick to int64_t and int32_t. It makes the API slightly more dangerous (using the int64_t methods from a 32-bit plat). But I think that's fine.

DrTimothyAldenDavis:
If we can get things to work using int32_t and int64_t, then it's less disruptive to the rest of SuiteSparse.

mmuetzel:
The signed type corresponding to the unsigned `size_t` would probably be `ptrdiff_t`.
A decision whether using that type would have been "better" than `int64_t` for 32-bit targets is probably better kept separate from the proposed changes here.
I agree that it might be best to use the same type here than for the other libraries (even if that might mean that half the API isn't really usable on 32-bit platforms).

DrTimothyAldenDavis:
Yes, if I had to start over, I would probably use a different strategy for integer sizes ... but I wrote some of these codes long before int32_t, int64_t, etc existed.

rayegun:
@DrTimothyAldenDavis things build now, but I'm slogging through a serious correctness issue (I'd bet I deleted a function call or something by accident. If you have a chance to run the tests and see anything that catches your eye let me know!

rayegun:
Nevermind I believe I found the issue

rayegun:
I don't love the number of explicit specializations, but I think it's fairly unavoidable. 

The only other thing I need to look at is adding int32_t tests and moving the cholmod wrappers into a header file rather than a .cpp file.

Oh and name bikeshedding for the int32_t functions (there are very few actually).

GPUQREngine/Source/BucketList/BucketList.cpp
```diff
@@ -105,13 +106,31 @@ BucketList::BucketList
...
+template BucketList<int32_t>::BucketList
+(
+    Front <int32_t> *F,
+    int32_t minApplyGranularity
+) ;
+template BucketList<int64_t>::BucketList
+(
+    Front <int64_t> *F,
+    int64_t minApplyGranularity
+) ;
```
mmuetzel, line 115: Is it really necessary to explicitly instantiate each member function? Wouldn't it be enough to just instantiate the class template?

I.e., after all (templated) member functions and possible specializations are defined, just add the following two lines to this file:
```
template class BucketList<int32_t>;
template class BucketList<int64_t>;
```

IIUC, you are still missing that explicit template class instantiation anyway. If you add that, you can probably remove the explicit instantiation of the class member functions.

GPUQREngine/Include/GPUQREngine_BucketList.hpp
mmuetzel, line 1: Afaict, the `extern template` declarations for the two instantiations that should be exported by the library are still missing from this file.

mmuetzel:
> Yes, if I had to start over, I would probably use a different strategy for integer sizes ... but I wrote some of these codes long before int32_t, int64_t, etc existed.

The old types (`int` and `long`) had "good" sizes on Linux. I.e., both are 32-bit wide on 32-platforms and `long` is 64-bit wide on 64-bit platforms. Windows made it a bit odd because `long` is 32-bit on 64-bit Windows. `long long` is needed for 64-bit integers on that platform.
With that approach, all functions from the SuiteSparse libraries could be used on 32-bit platforms.
With the new approach (`int32_t` and `int64_t`), the `*_l_*` functions from the API can no longer (reasonably) be used on 32-bit platforms.
It might have been better to use `ptrdiff_t` instead of `int64_t` in the `*_l_*` API. That would have "automatically selected" either `long` or `long long` on the respective platform.

rayegun:
The SPQR C API only has two different functions in this PR actually, I think that's a pretty minimal change and all the other *functions* will work as is, you just have to provide your cholmod_sparse structs in int32_t form.

I'll fix your comments above, but I think the API issue is probably fine. I *can* perhaps detect when cholmod_sparse structs with int64_t members are used on a 32-bit machine, but I'm not sure how I would do that in C++ yet.

In that case SPQR is pretty much identical to before, and you just have to ensure you don't use the `cholmod_l` methods on 32-bit architectures as before.

GPUQREngine/Source/BucketList/BucketList.cpp
rayegun: I ran into quite a few undefined ref errors at link time until I manually instantiated everything. I'll experiment a bit more, I'm unfortunately not an expert at templating.

mmuetzel:
> I _can_ perhaps detect when cholmod_sparse structs with int64_t members are used on a 32-bit machine, but I'm not sure how I would do that in C++ yet.

IIUC, the other libraries pushed the responsibility to use a functioning part of the API to downstream projects. It's probably fine if you do the same here.
I'm not sure that was the best option. But that shouldn't hold back this PR imho.

rayegun:
@DrTimothyAldenDavis this is more or less ready for an initial review.

My apologies for the gigantic size. There should be no logic changes, just a large reorganization to get templating working correctly.

If we can work towards getting this merged relatively quickly and released that would be super great. We have lots of users complaining that Julia cannot yet move to SS7 which is what most package managers now provide.

mmuetzel:
If I understand correctly, you designed to install many of the source files to allow implicit instantiations in all compilation units and in downstream code.
I don't know if that is the best approach.
I'd really recommend to try and use a combination of explicit template instantiations in one single compilation unit and `extern template` declarations in all others.

rayegun:
Where exactly would I put these extern template declarations? In each .cpp, except for the one chosen for explicit instantiation? 

Will it involve changing all the template files? That would be quite onerous, for what I measure to be a few seconds of extra compilation induced by this PR.

rayegun:
@DrTimothyAldenDavis this passes all of our 64-bit tests for Julia, so I'm going to set up a test of it for our 32-bit CI. This doesn't preclude the changes mentioned by @mmuetzel of course, once I understand what to do.

mmuetzel:
> Where exactly would I put these extern template declarations? In each .cpp, except for the one chosen for explicit instantiation?

You could do that. But it might be easier to do that in the header file with the template declarations that you'd need to include anyway.

> Will it involve changing all the template files?

It might be a good idea to revert the file extension change for those files (back from `.tpp` to `.cpp`). I haven't looked at them in much detail. But I'd expect that not much more than that should be needed.

> That would be quite onerous, for what I measure to be a few seconds of extra compilation induced by this PR.

It's not about compilation time. But sometimes weird effects can occur if a template is instantiated in multiple compilation units. Maybe some see some specializations that others don't (or they are optimized differently in different compilation units). That might not be an issue currently. But it might become one if one isn't very careful with future changes.


Edit: I tried to outline the rough approach in my first comment to this PR.

mmuetzel:
Maybe also check your editor settings.
It looks like it removed the linefeed at the end of many files.

rayegun:
> It might be a good idea to revert the file extension change for those files (back from .tpp to .cpp). I haven't looked at them in much detail. But I'd expect that not much more than that should be needed.

This goes against what I understand as the guideline to always put templates in headers if they may be used from more than one compilation unit? 

> make templates extern

I believe I understand how to do this, but it is an extraordinary amount of extra code for every templated function (since in SPQR they are all free functions). 

Since SuiteSparseQR is the entry point would it be sufficient to make that template extern?

rayegun:
I'll fix the newlines apologies

rayegun:
@mmuetzel is this last commit essentially what you're recommending? Would it be sufficient to do this for long functions and structs? Otherwise there is a lot of noise IMHO.

rayegun:
Okay someone has explained to me what is going on. I'll make the changes

mmuetzel:
> > It might be a good idea to revert the file extension change for those files (back from .tpp to .cpp). I haven't looked at them in much detail. But I'd expect that not much more than that should be needed.
> 
> This goes against what I understand as the guideline to always put templates in headers if they may be used from more than one compilation unit?

Afaict when it comes to templates, headers are meant for template (class) *declarations* (and maybe smaller definitions). They are meant to be included in other compilation units that use instantiations of the template (class).
The template (class) definition needs to have happened in a compilation unit before instantiating the template the first time. I'd really recommend instantiating the templates explicitly with all types that should be exported from the library or that are used in other compilation units.
To avoid (accidental) implicit instantiations in other compilation units where the header with the template (class) declaration is included, I'd recommend to add `extern template` declarations with all exported types at the bottom of that header.

I'm not sure if this is done consequently currently. I can try and have a closer look maybe some time the coming weekend.

All of that might be less problematic for templated functions. But template classes can become a little more confusing...


Edit: Sorry. I haven't come around to have a closer look on this during the weekend. And I'm not sure when I'll be able to.

rayegun:
@DrTimothyAldenDavis this compiles on my machine :/ I'm not sure what changed with the Complex typedef.

DrTimothyAldenDavis:
Maybe the C API and C++ API are conflicting.  C++ applications that include CXSparse cs.h are now using the C type.  Maybe cs.h and the my_cxx.h are colliding.

rayegun:
Ah I think it's because I now need the Complex type in SuiteSparseQR.hpp, whereas before it was only needed in spqr.hpp... I'm not sure how to solve that immediately.

mmuetzel:
The error message from clang++ is a bit clearer:
```
  In file included from /home/runner/work/SuiteSparse/SuiteSparse/Example/Source/my_cxx.cc:7:
  In file included from /home/runner/work/SuiteSparse/SuiteSparse/Example/Include/my_internal.h:30:
  /home/runner/work/SuiteSparse/SuiteSparse/include/SuiteSparseQR.hpp:29:30: error: typedef declarator cannot be qualified
  typedef std::complex<double> Complex ;
                               ^~~~~~~
  /home/runner/work/SuiteSparse/SuiteSparse/include/SuiteSparseQR_C.h:16:22: note: expanded from macro 'Complex'
  #define Complex std::complex<double>
                  ~~~~~^
```

So essentially, `SuiteSparseQR_C.h` is included somehow before `SuiteSparseQR.hpp`. The former defines `Complex` to `std::complex<double>`. So, the line it is complaining about essentially reads `typedef std::complex<double> std::complex<double> ;`.
To resolve this, it might help if you replace that line in `SuiteSparseQR_C.h` by `typedef std::complex<double> Complex ;`.

That's another instance where different types might be used in function interfaces depending on whether this header is included in C or C++ code. But the actual functions might be exported using the one or the other type (but not either interchangeably). That approach seems fragile. And the following comment might not be correct on all platforms:
```
/* The C++ functions will return a pointer to a std::complex<double> array of
   size n, which the C code must then interpret as double array of size 2*n,
   with real and imaginary parts interleaved. */
#define Complex double
```

Is that preprocessor macro needed at all? It isn't used in that header afaict. Can it just be removed?

GPUQREngine/Source/BucketList/BucketList.cpp
```diff
@@ -133,3 +133,5 @@ void BucketList::Initialize()
         }
     }
 }
+extern template class BucketList<int64_t>;
+extern template class BucketList<int32_t>;
```
mmuetzel: These `extern template` declarations at the end of a .cpp file can probably be removed.
Think of them like a forward declaration of a class that is defined in a different compilation unit.

What might make sense to do here is to explicitly instantiate the `template class` with those two types:
``` suggestion
template class BucketList<int64_t>;
template class BucketList<int32_t>;
```

The same for the other .cpp files.

GPUQREngine/Source/GPUQREngine_Internal.cpp
```diff
@@ -20,32 +20,40 @@
 #include "GPUQREngine_Scheduler.hpp"
 #include "GPUQREngine_Stats.hpp"
 
+template class BucketList<int32_t>;
+template class BucketList<int64_t>;
+template class LLBundle<int32_t>;
+template class LLBundle<int64_t>;
 
+template class Scheduler<int32_t>;
+template class Scheduler<int64_t>;
```
mmuetzel: Instead of explicitly instantiating the templates here, it might be better to "forward declare" them with `extern template` instead.
That reduces compilation time. But more importantly, it might also avoid accidentally skipping a specialization from another compilation unit in this compilation unit.

mmuetzel: And instead of having to remember to "forward declare" those template classes manually in each compilation unit, it might be easier to move those `extern template` declarations to the respective headers.

rayegun: If they are not instantiated there where do you intend to instantiate them? Implicitly? I was getting linker errors using implicit.

mmuetzel: Ah. Sorry. I realized only now that the member functions of these template classes are defined across several different compilation units. I only skimmed through the code. But afaict, there are no template specializations. (Please, check if that's true.)

If the entirety of a template class was defined in one single compilation unit, it would have made sense to explicitly instantiate these classes in those respective compilation units. But this project seems to be structured differently.

I'd guess that explicitly instantiating these template classes in this compilation unit should be ok in this case. 👍

Sorry for my confusion.

GPUQREngine/Source/BucketList/BucketList_CreateBundles.cpp
```diff
@@ -13,12 +13,11 @@
 //
 // =============================================================================
 
-#include "GPUQREngine_BucketList.hpp"
-
-
 // CreateBundles selects rowtiles up to PANELSIZE and creates a new bundle
 // ready for factorization.
-void BucketList::CreateBundles
+#include "GPUQREngine_BucketList.hpp"
```
mmuetzel, line 18: IIUC, the comment above this line belongs to the function below. Could please move the `#include` statement above again?

GPUQREngine/Source/BucketList/BucketList.cpp
```diff
@@ -21,23 +21,21 @@
...
     prev = (Int *) SuiteSparse_free(prev); \
     next = (Int *) SuiteSparse_free(next); \
     triu = (bool *) SuiteSparse_free(triu); \
-    Bundles = (LLBundle *) SuiteSparse_free(Bundles); \
+    Bundles = (LLBundle <Int> *) SuiteSparse_free(Bundles); \
     gpuVT = (double **) SuiteSparse_free(gpuVT); \
     wsMongoVT = Workspace::destroy(wsMongoVT);
-
-BucketList::BucketList
+#include "GPUQREngine_BucketList.hpp"
```
mmuetzel, line 34: Many projects prefer to have their `#include` as the first statements in a file if there is no particular reason for moving them down.
Could you please move this `#include` back up again?

GPUQREngine/Include/GPUQREngine_LLBundle.hpp
```diff
@@ -93,10 +117,14 @@ class LLBundle
...
+
     void UpdateSecondMinIndex();
     void UpdateMax();
+    
 
     void gpuPack(TaskDescriptor *cpuTask);
 };
 
+extern template class LLBundle<int32_t>;
+extern template class LLBundle<int64_t>;
```
mmuetzel, line 129: I'm not exactly sure if these `extern template` declarations will suffice to prevent implicit template instantiations from happening.
Some of the template class functions have arguments of another template class type (`BucketList<Int>`).
To be on the safe side, it might be better to also add `extern template` declarations for the two types of `BucketList` here.

Motivation: This file doesn't include `GPUQREngine_BucketList.hpp` but uses a forward declaration of that class template instead.

GPUQREngine/Source/TaskDescriptor/TaskDescriptor_flops.cpp
```diff
@@ -37,9 +37,9 @@ Int flopsFactorizeVT(int numTiles)
...
 // -----------------------------------------------------------------------------
 
-Int flopsFactorize(int m, int n)
+int64_t flopsFactorize(int m, int n)
 {
-    Int v = MIN(m, n);
+    int64_t v = MIN(m, n);
     return 2 * (m-1)                    +
            v * (6 + 4*m*n + 5*n + 2*m)  +
            ((-4*n -2*m -4) * v*(v+1)/2) +
```
mmuetzel, line 45: That might also have been an issue prior to this change:
Are these operations promoting the integer type to `int64_t` implicitly? ISTR that this only happens from smaller integer types to `int` (or `unsigned int`). It might be necessary to explicitly convert `m` and `n` to `int64_t` to get the expected result from these arithmetic operations if the values of `m` and `n` are very large.
Choosing an output type of `int64_t` suggests that the intention is to support large values.

mmuetzel:
Just a minor nit: When reading template types in C++ code, I find it easier to catch what belongs together if there is no space between the template name and the parameter list. E.g., `LLBundle<int64_t>` instead of `LLBundle <int64_t>` makes it clearer imho what belongs together to name the type.

But that comes down to personal preference and @DrTimothyAldenDavis should have the final say on style.

GPUQREngine/Source/TaskDescriptor/TaskDescriptor_flops.cpp
mmuetzel: Should be handled in #343.

DrTimothyAldenDavis: All those integers should probably be upcasted to int64_t first, just to be safe.  There might be other places where the computations should be done in int64_t, not in Int since the latter is int32_t with this update.  I'll need to go through carefully and check.

DrTimothyAldenDavis:
Regarding `LLBundle<int64_t>` versus  `LLBundle <int64_t>` :  I'm OK either way.

DrTimothyAldenDavis:
I'm about to post a stable v7.1.0 release of SuiteSparse (it wouldn't include this major change to SPQR).  Once I do that, I'll be ready to test this out.

Sound OK?  Is there anything I'm missing that I should add to the stable release of v7.1.0?

rayegun:
That sounds good to me. I'll go over formatting a bit tonight.

GPUQREngine/Source/compareTaskTime.cpp
mmuetzel, line 1: Is there a reason why this function has been moved to a separate compilation unit?
Wouldn't it even be possible to make that a file `static` function in its original location inside `Scheduler_TransferData.cpp`?

Edit: If it should really be moved to a new file, it'll probably need a copyright header.

DrTimothyAldenDavis:
I'm ready to merge this into dev2 (as well as the other related PRs: 
 https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/342 and https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/343

Is that OK?  Or shall I wait for more updates to this PR?

PR #318, opened 2023-05-16
CXSparse: Fix compilation in C++ mode.

mmuetzel:
This fixes a regression from #312 when the header is included in C++ code.
Two changes:
- Don't include header from the standard library in `extern "C"` block (mostly style).
- Use the standard `_Complex` specifier instead of the C-specific `complex`.

Sorry for not noticing earlier. (The example from #294 helped to catch that.)

### DIFF
--- a/CXSparse/Config/cs.h.in
+++ b/CXSparse/Config/cs.h.in
@@ -26,12 +26,13 @@
 #ifndef _CXS_H
 #define _CXS_H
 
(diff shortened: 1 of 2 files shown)


mmuetzel:
@DrTimothyAldenDavis: The scope of this change is pretty narrow. And it fixes a recently introduced regression (#312).
If you come around to it, could you please merge this?

PR #319, opened 2023-05-23
Add script to automatically build the library using Cmake VS generator.

flxh:
Currently the instructions for building the library under Windows suggest to import each submodule as a Cmake project into Visual Studio. I tried that, but it is kind of tiresome and prone to errors because the order in which you build the submodules matters and you have point Cmake to each of the previous build / install directories. 

This is an attempt to streamline this process using a simple powershell script that calls cmake, build and install using the Visual Studio Generator in Cmake. The submodules are build each in their own build directory build they are collectively installed to a common install path.

If this idea is of interest. I could extend the script to match the functionality of the Makefile even further.

### DIFF
--- a/Build_VS.ps1
+++ b/Build_VS.ps1
@@ -1,0 +1,26 @@
+$directories = @("SuiteSparse_config", "Mongoose", "AMD", "BTF", "CAMD", "CCOLAMD", "COLAMD", "CHOLMOD", "CSparse", "CXSparse", "KLU", "LDL", "UMFPACK", "SPQR")
+
+$blas_installation_path = "C:/OpenBLAS"
+$working_directory = (get-location).path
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
A combined cmake build system is in progress; I'm unsure if this will conflict with that (see the various open pull requests).

flxh:
> A combined cmake build system is in progress; I'm unsure if this will conflict with that (see the various open pull requests).

That's probably even better. Thank you for this continuing effort!

DrTimothyAldenDavis:
I'll close this now.  The open issue is here, where I'll resolve it with cmake: https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/175

PR #330, opened 2023-06-12
comment about building on Windows

DrTimothyAldenDavis:


mmuetzel:
It might be important to distinguish between MSVC and MinGW on Windows here. MinGW usually doesn't use Multi-Config generators (just like they aren't used frequently in the POSIX world). But they might be required for MSVC. (At least, I needed to resort to a Ninja Multi-Config generator for the CI using MSVC.)

Afaict, the extra flag for the `cmake --build` command is needed when selecting a Multi-Config generator during the configure step of Cmake. That might be the default if the used Cmake is built for MSVC (or Intel, or clang-cl, or ...).

PR #336, opened 2023-06-18
Clean-up for C++ example

mmuetzel:
Thank you for applying the other changes recently. 👍

This PR cleans up some interdependencies between those PRs.

Splitting the header in the example will probably allow to simplify the changes for #267 slightly.


DrTimothyAldenDavis:
looks great -- thanks!

DrTimothyAldenDavis:
I'd like to release a v7.1.0 of SuiteSparse soon, where #267 would appear in a later version (say v7.2.0).  I have a beta prelease of v7.1.0 already posted.  I can include these changes.  Are there others I should add?

mmuetzel:
Thanks for accepting this PR.

Imho, it might be good if the change from here could make it into v7.1.0. With them, the example is in a more "consistent" state imho.

#267 still needs some follow-up changes. Preferably in separate PRs. I agree that it would be better to wait until after the release for those changes.

DrTimothyAldenDavis:
Yes, that's what I thought.  I'll go ahead and include it in v7.1.0.beta2.  Thanks.

PR #340, opened 2023-06-21
Generated files should not be tracked in the repository

kiwifb:
While working on some other issues I noticed that after running cmake the timestamp of `SuiteSparse_config.h` appears untouched which lead me to wonder if it was regenerated at all from `Config/SuiteSparse_config.h.in` by the cmake configuration process. And of course in my case removing that file to force regeneration lead to an identical file, is cmake trying to be smart or is it a genuine issue? In any case a file generated from a template during configuration should not be tracked in the repository. So removing it and adding `.gitignore` instructions for it feels the right thing to do.


mmuetzel:
I agree that it might make it easier if those generated files weren't tracked. In many cases, the local clone was marked as modified for me just because I was testing with a different compiler, or something similar caused the generated files to be slightly different. That makes selecting the correct files for a commit slightly more tedious.

But ISTR that @DrTimothyAldenDavis wrote somewhere else that he prefers those files to be always present.

In any case, if this change is accepted (which I personally would be in favor of), it would probably make sense to remove all of those generated files (not just this single one).

DrTimothyAldenDavis:
I still have some Makefiles that don't use cmake, and they need access to those generated files.  I plan on revising them so I won't need to keep the generated files in github but for now I need to leave them in.

kiwifb:
Fair enough.

DrTimothyAldenDavis:
I do agree that it is a bit confusing to see files changed from the perspective of git.

But to fix this, I would also have to figure out how to do without the configured *.h files when the libraries are compiled solely for use inside MATLAB.  That uses the mex command inside MATLAB itself.  It is hard to use cmake are even a mere Makefile.  The configured *.h files typically have defaults that can be over-ridden by the m-file scripts using -D options (such as how to call the Fortran BLAS, which is different inside MATLAB).  So they can use the *.h files that I configure and post in github (assuming no .gitignore) on any platform.

These m-file scripts would thus have to be changed to create all the *.h files.  All of my packages have at least one configured *.h file, like amd.h for AMD.  It includes a configured version number so I don't have to edit the version number in more than one place.

For all these reasons it would be challenging to use .gitignore to remove all configured *.h files in SuiteSparse.

PR #341, opened 2023-06-23
Set the required C/C++ standard in more cmake rules

mmuetzel:
I missed a couple of `CMakeLists.txt` files in subfolders as part of #295. Here, they are. 🙂


DrTimothyAldenDavis:
Yes -- those should be updated too.

PR #342, opened 2023-06-25
SuiteSparseQR_C.h: Don't define `Complex`

mmuetzel:
The preprocessor macro `Complex` is defined to types with possibly different alignment and size depending on whether this header is included in C or C++ code. It's defined to `std::complex<double>` when included in C++ code and `double` when included in C code. This could lead to issues if that macro is used for function arguments that are used in C or C++ code.
That macro doesn't seem to be used anywhere in SPQR itself currently. So, just remove it. 

(I don't know whether it's used in any downstream project though. But `double _Complex` might be the "better" type anyway because the standard guarantees alignment and size that is compatible to `std::complex<double>` afaict. But the ABI might still be incompatible to `std::complex<double>`...)

Also, avoid including headers inside an `extern "C"` block.


DrTimothyAldenDavis:
Yes, this would be better.  I don't think I use this `#define` in other packages but I'll check.  I should probably include this in SuiteSparse v7.2.0, however, because it is a user-visible change.  It would bump the major SO number for SPQR.  The changes by @Wimmerer will also likely bump the major SO number of SPQR as well, so I'd like to do both at the same time.

I think v7.1.0 is ready to post as a stable release but I'll check the recent PR's just to be sure.

DrTimothyAldenDavis:
This should go to the dev2 branch, not the dev branch.

mmuetzel:
> This should go to the dev2 branch, not the dev branch.

Oops. I must have forgotten to change that when creating the PR (even if the default PR message says what to do). The default branch (`dev`) is always pre-selected. We needed that branch as the default branch for the GitHub cache behavior.

Would it make sense to activate the CI on `dev2`? Imho, it doesn't matter if the CI would be failing once in a while.

If that is done, the default branch could be set to be `dev2`. That would pre-select the "correct" branch when opening a PR. (I also noticed a couple of other PRs opened against `dev`.)

DrTimothyAldenDavis:
I like the dev branch to be the default. I should be the only one to push to it since it's the default. And I should do a push there only from the dev2 branch. That's also where all other PRs from everyone else should go. 

Whatever the default branch is, I see it as a fairly stable beta branch with pending updates that will appear in the next stable release.

The dev2 branch can be broken and have code in progress. 

So let's leave things as they are. If a PR comes in to the stable or dev branches, I'll just rebase it to dev2

PR #343, opened 2023-06-26
GPUQREngine: Avoid possible integer overflow in arithmetic operations

mmuetzel:
For some input values, some arithmetic operations in `TaskDescriptor_flops.cpp` might overflow when they are performed with the input type `int`. Convert to the output type (corresponding to `int64_t`) before those operations to move those potential integer overflow issues to much larger values.

### DIFF
--- a/GPUQREngine/Source/TaskDescriptor/TaskDescriptor_flops.cpp
+++ b/GPUQREngine/Source/TaskDescriptor/TaskDescriptor_flops.cpp
@@ -2,7 +2,7 @@
 // === GPUQREngine/Source/TaskDescriptor_flops.cpp =============================
 // =============================================================================
 
-// GPUQREngine, Copyright (c) 2013, Timothy A Davis, Sencer Nuri Yeralan,
+// GPUQREngine, Copyright (c) 2013-2023, Timothy A Davis, Sencer Nuri Yeralan,
 // and Sanjay Ranka.  All Rights Reserved.
 // SPDX-License-Identifier: GPL-2.0+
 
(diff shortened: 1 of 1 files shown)


mmuetzel:
Moving a comment over from #316:
https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/316#discussion_r1244141779
@DrTimothyAldenDavis wrote:
> All those integers should probably be upcasted to int64_t first, just to be safe. There might be other places where the computations should be done in int64_t, not in Int since the latter is int32_t with this update. I'll need to go through carefully and check.

Which particular operation are you worried about? Afaict, all pairs of operands will have at least one member which has the size of the final output. So, we should be fine afaict.

IIUC, `Int` is currently `int64_t`. In any case, it should be sufficient to do the operations in the precision that is required for the output. It shouldn't be necessary to do the operations in a higher precision as the one used for the result (on the vast majority of platforms at least). Or did I misunderstand your comment?

mmuetzel:
The calculation in `flopsFactorizeVT` and `flopsFactorize` look pretty similar apart from a factor `2` vs. `3` in the 4th term.
Is that a typo or is that intentional? In case it is a typo: which one is correct?

DrTimothyAldenDavis:
All of these methods should typecast their inputs from int or Int into int64_t, do all their work in int64_t, and then they should all return int64_t, not Int.  I can take care of that.

DrTimothyAldenDavis:
Thanks for catching this issue.  I'll close this PR since I'm working on a different approach.  There are other issues with flop count overflow in this int32_t version of SPQR.

PR #351, opened 2023-07-04
CXSparse: Avoid including C header in C++ code

mmuetzel:
According to [1]: "Source files that are not intended to also be valid ISO C should not use any of the C headers."

It is still correct to define `cs_complex_t` as `double _Complex` in C and C++ code. But the C standard header `complex.h` shouldn't be included in a header that can be used in C++ code. That is not a requirement. But it is a recommendation.
I wasn't aware of that recommendation for PR #312.

Additionally, avoid inclusion of headers inside an `extern "C"` block.

[1]: https://en.cppreference.com/w/cpp/standard_library

### DIFF
--- a/CXSparse/Config/cs.h.in
+++ b/CXSparse/Config/cs.h.in
@@ -27,12 +27,12 @@
 #define _CXS_H
 
 #if @CXSPARSE_USE_COMPLEX@
-#include <complex.h>
-#define cs_complex_t double _Complex
-#endif
-
-#ifdef __cplusplus
-extern "C" {
+#  ifdef __cplusplus
+#    include <complex>
+#  else
+#    include <complex.h>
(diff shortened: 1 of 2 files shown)


mmuetzel:
I'm not sure what up with homebrew recently. Are we using it incorrectly?

Edit: That is failing currently probably because of https://github.com/Homebrew/homebrew-cask/issues/150323.

mmuetzel:
Rebased on a current head of the dev2 branch.

@DrTimothyAldenDavis: This is a trivial change for standard conformance. Can this be merged?

DrTimothyAldenDavis:
Looks good.  Still getting caught up from several weeks of travel, and now some kind of viral fatigue, so I'll still be slow this next week.

PR #353, opened 2023-07-21
CI: Add Alpine Linux runners on different (emulated) CPU architectures

mmuetzel:
In the light of #221, this PR adds some additional runners to the CI that build and run the SuiteSparse libraries on different CPU architectures.

There is an existing action that makes it pretty easy to install a chroot for different CPU architectures based on the packages distributed by Alpine Linux. That action also allows it to easily execute commands "natively" on those architectures. If necessary, they are emulated by qemu.

I tried to keep the rules pretty close to the existing CI. But building GraphBLAS on emulated hardware takes too long. In some prior tests, the CI didn't complete even 10% of the compilation process before the 6 hour timeout was reached.
So, I skipped building GraphBLAS for these runners. To still be able to build the example, this PR adapts it such that GraphBLAS is now an *optional* dependency for the example. I hope that is ok.

As you can see, some of the demos segfault on CPUs for which the pointer type is 32-bit. We already saw the same for Windows 32-bit. But it looks like it is also happening on Linux.


mmuetzel:
Rebased on a current head of the dev2 branch to check if this makes a difference for the platforms here (especially the ones with 32-bit pointers).

mmuetzel:
It looks like the tests for SPQR no longer crash on 32-bit platforms.

@DrTimothyAldenDavis: Do you think it would make sense to add these rules to the CI? I.e., could you please merge this PR?

DrTimothyAldenDavis:
Yes -- this is a great idea.  I've gotten behind because of travel this past month, and I came back from my last trip with a cold.  I should be able to get caught up this week.

DrTimothyAldenDavis:
These changes look fine.  Making GraphBLAS an optional part of the Example package is OK.  An alternative would be to compile GraphBLAS with its COMPACT option turned on.

For now, I'll go ahead and merge this in and give it a try.

PR #356, opened 2023-08-15
Create pkg-config files

mmuetzel:
The CMake modules make it pretty easy to link to any of the SuiteSparse libraries without much knowledge about the host system or dependencies.
Projects not using CMake currently need to find and track the libraries and dependencies manually. For those projects, it would be helpful if SuiteSparse installed pkg-config files which can be used to query compiler flags and linker flags easily across multiple platforms.

This PR adds rules to the CMake files to generate such pkg-config files for the libraries that are built as part of SuiteSparse.

E.g.:
```
$ pkg-config --cflags CHOLMOD
-I/usr/include
$ pkg-config --libs CHOLMOD
-L/usr/lib -lcholmod
```

If the package config files are not installed in a default location, `PKG_CONFIG_PATH` can be set to point to the correct one. E.g.:
```
$ PKG_CONFIG_PATH=~/SuiteSparse/lib/pkgconfig/ pkg-config --libs --static CHOLMOD
-L~/SuiteSparse/lib -lcholmod -lgomp -lthread -lthread -lm -lopenblas -lopenblas -lopenblas -L~/SuiteSparse/lib -lccolamd -lcamd -lcolamd -lamd -lsuitesparseconfig
```
(Note that the last command prints the linker flags for *static* linking.)

I don't know how to use CUDA. So, I don't know which flags need to be added for that. If users would like to link to CUDA, they'd still need to find the correct flags themselves. I left *FIXME* comments in the .pc files of the respective libraries.

The CMake function `cmake_path` was added in version 3.20 according to their documentation. Some CMake files already require version 3.22. For the others, I increased the minimum version to 3.20 (released March 2021).

The `Description` tag could probably be improved. I tried to use one of the first sentences of the documentation of each library to compose a short description and added "in SuiteSparse" to the end. I hope that is ok for a start.

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -10,7 +10,7 @@
 # get the version
 #-------------------------------------------------------------------------------
 
-cmake_minimum_required ( VERSION 3.19 )
+cmake_minimum_required ( VERSION 3.20 )
 
 set ( AMD_DATE "June 16, 2023" )
 set ( AMD_VERSION_MAJOR 3 )
@@ -145,6 +145,7 @@ endif ( )
 if ( NOT WIN32 )
     target_link_libraries ( AMD PRIVATE m )
     if ( NOT NSTATIC )
+        set ( AMD_STATIC_LIBS "${AMD_STATIC_LIBS} -lm" )
         target_link_libraries ( AMD_static PUBLIC m )
     endif ( )
 endif ( )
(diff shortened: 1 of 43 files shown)


DrTimothyAldenDavis:
The pkg-config idea is new to me and I don't understand how to resolve these merge conflicts.

mmuetzel:
Rebased on a current head of the `dev2` branch to resolve some conflicts.

.pc files pre-date CMake and are commonly used (especially on Linux or macOS) to facilitate getting build and link flags to include a project downstream.
https://en.wikipedia.org/wiki/Pkg-config

They can be used by any build system -- including autotools/GNU make or CMake. Essentially, they can be used from anywhere that can run the `pkg-config` binary.

DrTimothyAldenDavis:
Thanks!  I'll give this a try early next week.

DrTimothyAldenDavis:
I'll merge this after I get the link errors with CUDA::nvrtc fixed.  Hopefully soon.  Thanks!

PR #358, opened 2023-08-29
CMake: Import CUDA toolkit targets when importing CHOLMOD_CUDA.

mmuetzel:
Also don't require that consumers of the shared libraries need to link to CUDA.


DrTimothyAldenDavis:
Yes, that would be better.  I'll give it a try.

Note that I also just now updated all of the version numbers in the CMakeLists.txt, and added a blurb to each PACKAGE/Doc/Changelog file.  This PR and #359 have no conflicts with those changes, fortunately.

PR #359, opened 2023-08-29
KLU: Don't explicitly link to transient targets.

mmuetzel:
`SuiteSparse::CHOLMOD_CUDA` is a transient target of `SuiteSparse::CHOLMOD`.
Don't link to it explicitly.

### DIFF
--- a/KLU/CMakeLists.txt
+++ b/KLU/CMakeLists.txt
@@ -296,43 +296,20 @@ if ( NOT NCHOLMOD )
 
     # CHOLMOD:
     # link with CHOLMOD and its dependencies, both required and optional
-
-    if ( SUITESPARSE_CUDA )
-
-        # CHOLMOD with CUDA
-        target_link_libraries ( KLU PRIVATE
-            SuiteSparse::CHOLMOD SuiteSparse::CHOLMOD_CUDA )
-        target_link_libraries ( KLU_CHOLMOD PRIVATE
-            SuiteSparse::CHOLMOD SuiteSparse::CHOLMOD_CUDA )
-        if ( NOT NSTATIC )
-           set ( KLU_STATIC_MODULES "${KLU_STATIC_MODULES} CHOLMOD CHOLMOD_CUDA" )
-           if ( TARGET SuiteSparse::CHOLMOD_static )
-                target_link_libraries ( KLU_static PUBLIC SuiteSparse::CHOLMOD_static )
-                target_link_libraries ( KLU_CHOLMOD_static PUBLIC SuiteSparse::CHOLMOD_static )
-            else ( )
-                target_link_libraries ( KLU_static PUBLIC SuiteSparse::CHOLMOD )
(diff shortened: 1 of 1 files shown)


mmuetzel:
It might be that this depends on #358 to work when building with CUDA.

DrTimothyAldenDavis:
I'm trying this out (and also #358) with CUDA now.

DrTimothyAldenDavis:
No luck.  FindCUDAToolkit isn't found.

```
CMake Error at /home/faculty/d/davis/dev2/SuiteSparse/CHOLMOD/build/CHOLMOD_CUDAConfig.cmake:63 (FindCUDAToolkit):
  Unknown CMake command "FindCUDAToolkit".
  /home/faculty/d/davis/dev2/SuiteSparse/CHOLMOD/build/CHOLMODConfig.cmake:65 (find_package)
  CMakeLists.txt:83 (find_package)
```

mmuetzel:
Did I mistype? It should be there:
https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html

DrTimothyAldenDavis:
I think the usage is

include ( FindCUDAToolkit )

instead of 

FindCUDAToolkit ( ... )

DrTimothyAldenDavis:
See GraphBLAS/cmake_modules/SuiteSparsePolicy.cmake and SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake where I use that format already.

DrTimothyAldenDavis:
I think this works:

`find_package ( CUDAToolkit @CUDAToolkit_VERSION_MAJOR@ REQUIRED )`

DrTimothyAldenDavis:
Oh.  Not quite.  That gets CHOLMOD to compile but then KLU fails:

```
[ 96%] Building C object CMakeFiles/KLU_CHOLMOD.dir/User/klu_cholmod.c.o
In file included from /home/faculty/d/davis/dev2/SuiteSparse/KLU/User/klu_cholmod.c:23:
/home/faculty/d/davis/dev2/SuiteSparse/CHOLMOD/Include/cholmod.h:456:10: fatal error: cublas_v2.h: No such file or directory
  456 | #include <cublas_v2.h>
      |          ^~~~~~~~~~~~~
```

mmuetzel:
> I think this works:
> 
> `find_package ( CUDAToolkit @CUDAToolkit_VERSION_MAJOR@ REQUIRED )`

Yes, that should do it.

I'm having doubts about the `REQUIRED`. It's only actually required when trying to use the static library. When linking to the shared library, users should only need to install the shared CUDA libraries. No need to install the dev package...

mmuetzel:
> Oh. Not quite. That gets CHOLMOD to compile but then KLU fails:

Ah. Didn't see that it was included in a header. So the `-I` flags should be PUBLIC. I'll look into it.

DrTimothyAldenDavis:
See my latest commit.  I fixed the FindCUDAToolKit but not the KLU failure.

mmuetzel:
#360 might help to find that header in dependent projects.

PR #360, opened 2023-08-29
CHOLMOD: Export include flags from cuBLAS.

mmuetzel:
When it is built with CUDA, consumers of the `SuiteSparse::CHOLMOD` target should use preprocessor flags to get the cuBLAS header.


DrTimothyAldenDavis:
I'm trying it now -- thanks.

DrTimothyAldenDavis:
OK -- that looks like it's working now. KLU compiles just fine.

I haven't tried building without CUDA yet but I figure that will work.

mmuetzel:
Building without CUDA works here at least.

PR #363, opened 2023-08-29
SPQRGPU: Respect local installation target.

mmuetzel:
I've been testing with `make CMAKE_OPTIONS="-DCMAKE_INSTALL_PREFIX=../.."`. So, `local` didn't make a difference for me.

Should be fixed with this change hopefully.


mmuetzel:
I grepped the source tree for `CMAKE_INSTALL_LIBDIR`. All remaining instances are fine or in parts that I don't know.
That is:
* GraphBLAS/cpu_features/CMakeLists.txt
* GraphBLAS/cpu_features/ndk_compat/CMakeLists.txt

DrTimothyAldenDavis:
Trying it now.

PR #364, opened 2023-08-30
CMake: Use `SUITESPARSE_INCLUDEDIR` in more places.

mmuetzel:
Also, don't double quote generator expressions. (I don't recall why I did that in the first place...)

The remaining place where `CMAKE_INSTALL_INCLUDEDIR` is still used is in `GraphBLAS/cpu_features/CMakeLists.txt`. I'm not sure what that is and left it alone.

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -98,8 +98,8 @@ set_target_properties ( AMD PROPERTIES
     WINDOWS_EXPORT_ALL_SYMBOLS ON )
 
 target_include_directories ( AMD
-    INTERFACE "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/Include>"
-              "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>" )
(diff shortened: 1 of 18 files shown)


DrTimothyAldenDavis:
The cpu_features folder is an exact copy of the original Google repo for that package, but I don't use the cpu_features build system itself to create a cpu_features library.  I compile the code from the GraphBLAS cmake.  So the cpu_features/CMakeLists.txt file is unused.  Same thing with CHOLMOD/SuiteSparse_metis/CMakelists.txt

mmuetzel:
> So the cpu_features/CMakeLists.txt file is unused.

In that case, the changes from this PR should be complete when it comes to the include directory in the interface of the CMake targets.

DrTimothyAldenDavis:
Looks good.

Removing the extraneous double quotes is a good idea.  I was confused by them since the spurious report from SuiteSparse_GPURuntime had no double quotes but the other places had it, so I thought that error was from missing a required double quote.

DrTimothyAldenDavis:
Works great -- thanks!  The Example package now compiles on my desktop when using CUDA.

PR #365, opened 2023-08-30
CHOLMOD: Don't look for CHOLMOD_CUDA in common source tree when installed

mmuetzel:
Create two slightly different `CHOLMODConfig.cmake` files. One that is used in a common build tree before CHOLMOD was installed. And a slightly different one that is installed.
The second one doesn't attempt to look for CHOLMOD_CUDA in the `build` folder of a common source tree.

### DIFF
--- a/CHOLMOD/CMakeLists.txt
+++ b/CHOLMOD/CMakeLists.txt
@@ -510,9 +510,18 @@ install ( EXPORT CHOLMODTargets
     NAMESPACE SuiteSparse::
     DESTINATION ${SUITESPARSE_LIBDIR}/cmake/CHOLMOD )
 
+# generate config file to be used in common build tree
+set ( SUITESPARSE_IN_BUILD_TREE on )
 configure_package_config_file (
     Config/CHOLMODConfig.cmake.in
     ${CMAKE_CURRENT_BINARY_DIR}/CHOLMODConfig.cmake
+    INSTALL_DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/CHOLMODConfig.cmake )
+
+# generate config file to be installed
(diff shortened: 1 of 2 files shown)


DrTimothyAldenDavis:
Would the same approach need to be done for all other *Config.cmake files?

mmuetzel:
> Would the same approach need to be done for all other *Config.cmake files?

Most of the *Config.cmake files don't automatically look for more targets than themselves. (Maybe they should?)
The one for SPQR was one of them. But that should be fixed by a part of #369.
I'm going to check all others whether they need the same treatment when I come around to it.

mmuetzel:
> I'm going to check all others whether they need the same treatment when I come around to it.

The only *Config.cmake files that are looking for other SuiteSparse libraries "by themselves" are the ones for the CHOLMOD libraries and SPQR.
These have already been updated to contain the logic to only look for the `build` folders in a common source tree *before* they are installed.

DrTimothyAldenDavis:
That makes sense.

I'm fine with the existing structure.  A user's package that uses UMFPACK, for example, would need to do

```
find_package ( AMD  ...)
find_package ( COLAMD ...)
find_package (UMFPACK ...)
```

and so on.  That's fine with me.  It's the optional CUDA libraries that CHOLMOD and SPQR require that should be hidden from the user project cmake.

mmuetzel:
Ideally, a user shouldn't need to know that they need to import the AMD and COLAMD packages if they exclusively used functions from UMFPACK.
But I agree that this shouldn't block a release.

DrTimothyAldenDavis:
Agreed.  And that could be the case even if they use AMD.  Just importing UMFPACK could ensure that AMD is imported.  But it's fine to leave as-is.

PR #367, opened 2023-08-30
CHOLMOD GPU: Fix openmp

whuaegeanse:
On windows, MSVC throws openmp related compilation errors

### DIFF
--- a/CHOLMOD/GPU/t_cholmod_gpu.c
+++ b/CHOLMOD/GPU/t_cholmod_gpu.c
@@ -49,7 +49,7 @@ void TEMPLATE2 (CHOLMOD (gpu_clear_memory))
     int chunk_multiplier = 5;
     int num_chunks = chunk_multiplier * num_threads;
     size_t chunksize = size / num_chunks;
-    size_t i;
+    int i;
 
 #pragma omp parallel for num_threads(num_threads) private(i) schedule(dynamic)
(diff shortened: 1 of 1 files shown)


mmuetzel:
I can only guess which error you are seeing. But would it be better to make `num_chunks` a `size_t` instead of making `i` an `int`?
That might help with integer overflow for large numbers.

whuaegeanse:
> I can only guess which error you are seeing. But would it be better to make `num_chunks` a `size_t` instead of making `i` an `int`? That might help with integer overflow for large numbers.

@mmuetzel 
MSVC Error:
`2>D:\zgh\Libs\suitesparse\SuiteSparse\CHOLMOD\Supernodal\../GPU/t_cholmod_gpu.c(55,9): error C3016: 'i': index variable in OpenMP 'for' statement must have signed integral type`

mmuetzel:
Thank you for testing.
See also this similar issue where it seems to have been solved by adding `/openmp:llvm` to the compiler flags:
https://github.com/tesseract-ocr/tesseract/issues/3285

Does that also work for you?

mmuetzel:
Looking at the affected code snippet again: `num_chunks` is already of type `int`. So, `i` can be of the same size.

Your change looks good to me. 👍

DrTimothyAldenDavis:
Looks good -- thanks.

PR #371, opened 2023-08-31
CMake: Use `SUITESPARSE_INCLUDEDIR` in even more places.

mmuetzel:
Also, don't double quote generator expressions.

Those are some instances that I missed in #364.

I'm pretty sure that I grepped for `CMAKE_INSTALL_INCLUDEDIR` when I made those changes. But my concentration might have been low (late evening here)…


DrTimothyAldenDavis:
Thanks -- I'll give it a try.

PR #376, opened 2023-09-01
CSparse: Don't install pkg-config file.

mmuetzel:
Nothing else is installed for CSparse. So, don't install the pkg-config file either.

Should it be built at all with the default rules in the root Makefile if it doesn't get installed anyway?

### DIFF
--- a/CSparse/CMakeLists.txt
+++ b/CSparse/CMakeLists.txt
@@ -149,9 +149,9 @@ configure_file (
     CSparse.pc
     @ONLY
     NEWLINE_STYLE LF )
-install ( FILES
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
CSparse should still be built, just not installed (see the top-level Makefile where "make" will compile CSparse, but "make install" will not install it).  

CSparse is the code in my book, https://epubs.siam.org/doi/book/10.1137/1.9780898718881 .  It's meant for educational use, so I would want students to use it in classes.  CSparse is production-quality, but it doesn't include the complex versions, which are in CXSparse.  CXSparse is meant for production use.  It actually has the same name of the header file, cs.h, as CSparse.

That's why CSparse should normally not be installed.

PR #377, opened 2023-09-03
CHOLMOD: Use CMake default functions to check for type existence.

mmuetzel:
Also remove CMake module that is no longer being used.


DrTimothyAldenDavis:
Trying it now -- thanks.

PR #378, opened 2023-09-04
CI: Add runners that build with CUDA on Ubuntu

mmuetzel:
This extends the matrix for the Ubuntu runners to include builds with and without CUDA.

I hope that will be useful in testing.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,12 +21,15 @@ jobs:
     # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
     runs-on: ubuntu-latest
 
+    name: ubuntu (${{ matrix.compiler }} ${{ matrix.cuda }} CUDA)
+
     strategy:
       # Allow other runners in the matrix to continue if some fail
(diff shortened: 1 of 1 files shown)


mmuetzel:
The workaround for `nvlink` is needed because those three libraries are empty on Ubuntu. ` nvlink` seems to require that a static library consists of at least one object file (even if that object file is empty).
According to internet sources, this is fixed in newer versions of `nvlink`. Version 11.7 is reported to work correctly with those original empty libraries. But Ubuntu 22.04 distributes version 11.5.

Maybe, that work-around can be removed when the CI on GitHub switches to Ubuntu 24.04 (probably some time next year).

DrTimothyAldenDavis:
Awesome -- thanks.  I'll merge it as soon as the checks complete.

DrTimothyAldenDavis:
I'm currently working on SPQR, by adding the test coverage suite for 32-bit integers added by @Wimmerer.  A few changes to CHOLMOD to support that update, as well.  Those changes shouldn't conflict with these changes to the build.yaml.

PR #379, opened 2023-09-05
CI: Use ccache also for CUDA compiler.

mmuetzel:
Forgot about that in #378.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -50,6 +50,7 @@ jobs:
             cuda-cmake-flags:
               -DENABLE_CUDA=On
               -DCUDAToolkit_INCLUDE_DIR="/usr/include"
+              -DCMAKE_CUDA_COMPILER_LAUNCHER="ccache"
 
     env:
       CC: ${{ matrix.cc }}


DrTimothyAldenDavis:
Looks good.  I'll merge it in once the CI passes.

I'd like to release SuiteSparse v7.2.0 soon, with all these changes.  I think things are working pretty well now with the build system, and with my revisions to the 32-bit SPQR by @Wimmerer .

Does this sound OK to both of you, @mmuetzel and @Wimmerer ?

I'll post it as a beta version first, to get more feedback.  I'll also post a v8.2.0.beta on my GraphBLAS repo. Then v9.0.0 of GraphBLAS will be posted soon on my GraphBLAS repo, since it implements the upcoming v2.1 C API for GraphBLAS (to appear sometime in September).

mmuetzel:
> I'd like to release SuiteSparse v7.2.0 soon, with all these changes.

Sounds good to me.

I've a couple of incremental changes in the pipeline that could simplify the CMake rules a bit now that we are using CMake targets. But none of them are worth delaying a release.

The only thing that might be nice to have fixed before a release is probably #221. That still means that 32-bit binaries are stuck at SuiteSparse 5.

rayegun:
32-bit binaries are only stuck if unwilling to update to the new functions right? A major version release is breaking, and there have been several since v5.

mmuetzel:
> 32-bit binaries are only stuck if unwilling to update to the new functions right? A major version release is breaking, and there have been several since v5.

Did you read my comment there?
Citing here for convenience:
> I also tried cherry-picking this change to SuiteSparse 7.1.0. With it, Octave no longer segfaults during the test suite. But I still see a failing test that used to pass with SuiteSparse 5:
> 
> ```
> >>>>> processing C:\msys64\mingw32\share\octave\8.2.0\etc\tests\libinterp\corefcn\qr.cc-tst
> ***** testif ; (__have_feature__ ("SPQR") && __have_feature__ ("CHOLMOD")) || __have_feature__ ("CXSPARSE") <*63069>
>  assert (qr (sparse (0, 0)), sparse (0, 0))
>  assert (qr (sparse (1, 0)), sparse (1, 0))
>  assert (qr (sparse (0, 1)), sparse (0, 1))
> !!!!! regression: https://octave.org/testfailure/?63069
> integer dimension or index out of range for Octave's indexing type
> ```
> 
> The last message could indicate that some indices or sizes might still be returned with bogus values.

I'll try again once the beta has been tagged.

DrTimothyAldenDavis:
I'm not sure what to make of that test failure in Octave.  It could be that the 32-bit Octave is still trying to call the 64-bit SPQR by mistake.  I don't think that should hold up a beta release of SuiteSparse v7.2.0 (with SPQR v4.2.0 beta, including the new option for 32-bit indices).

A beta release would be nice to check again with Octave, anyway.

DrTimothyAldenDavis:
I will tag a beta release later today, after this PR is merged.  I'm also doing my exhaustive tests ("make cov", and such) that aren't in the github CI process.  Once those all pass on my 4 platforms (linux with GPU, linux no GPU, Mac Arm, Intel Mac), I'll tag SuiteSparse v7.2.0.beta and also GraphBLAS v8.2.0.beta on the GraphBLAS repo.

mmuetzel:
> It could be that the 32-bit Octave is still trying to call the 64-bit SPQR by mistake.

Would that not be a supported use case if Octave converts everything to and from 64-bit when interfacing SPQR? That worked in SuiteSparse 5.
I thought the whole reason for changing `SuiteSparse_long` was to make sure that it also works on 32-bit platforms...

DrTimothyAldenDavis:
Yes, that would be supported.  But maybe Octave is calling the wrong integer version of SPQR, since now there are two of them.  It would need to call `spqr_whatever <double,int32_t> (parameters here).`

mmuetzel:
> It would need to call `spqr_whatever <double,int32_t> (parameters here).`

Why? Are 32-bit application forbidden to call `spqr_whatever <double,int64_t>`? Then why was `SuiteSparse_long` changed at all? That change only affected 32-bit platforms...

mmuetzel:
I hope I'm just misunderstanding and you are not writing: "Half of the API is broken on 32-bit platforms. But that was intentional. Just use the other half."
(Because it sounds a bit like that...)

DrTimothyAldenDavis:
Sorry for the confusion.

They can call either version. Both work on a 32 bit platform

They can even probably call it with ptrdifft. And it would pick the right
one

mmuetzel:
Thanks for clarifying. So, how could that error be caused by Octave calling the wrong integer version of SPQR?

DrTimothyAldenDavis:
Say Octave has an integer type that denotes its index type; call it idx_t.  On 32-bit platorms, idx_t is int32, and on 64-bit platforms, it's int64 (that's not required but say it does that).

Then Octave should call spqr_whatever <double,idx_t> (...).

What might be happening is the 32-bit Octave is calling spqr_whatever<double>(...), expecting to get the 32-bit integer variant of SPQR (the only one in SuiteSparse 5 if compiled on a 32-bit platform).  But now it gets int64, and it's confused.

That's just entirely hypothetical.  That error would cause a catastrophic failure in Octave, however, not just a single test.

But maybe that test is triggering something like this, just in one particular case.

Just a theory.  I'd need to look at the Octave interface to SPQR to figure out what's going on.

Either way, Octave now has the freedom to handle all 4 cases: (int32 or int64 indices) * (32-bit or 64-bit platform).

mmuetzel:
To stay compatible with earlier versions of SPQR (that's a requirement that we can't give up), Octave converts all integers between its own indexing type and `SuiteSparse_long` when interfacing to SPQR. So, it is now effectively using the `int64_t` interface (independent on the platform).

You seem to be convinced that there is an error in Octave. But by default, Octave's indexing type used to be 32 bits wide independent on the platform. And that is working correctly on 64-bit platforms for years where the only SPQR had only a 64-bit interface for the longest time.
So, it's pretty unlikely that some part of this conversion is missing. (Although not impossible.)

The changes in SPQR are more recent. So, I'd guess that it is more likely that something is still not doing the correct thing on that end.

PR #380, opened 2023-09-05
SuiteSparse v7.2.0.beta1

DrTimothyAldenDavis:
Sept 8, 2023: version 7.2.0 (BETA1)

    * build system: modern cmake structure, by Markus Muetzel.
        Most packages updated to vX.2.0 where X is unchanged (except SPQR
        and Example package).
    * SPQR v4.2.0: Major SO update.  Support for int32 indices by Raye Kimmerer


mmuetzel:
There seems to be a build issue when CHOLMOD is configured with `NOPENMP`.
I guess that moving `SuiteSparse__thread.cmake` to the `cmake_modules` folder inside `SuiteSparse_config` might fix it. But I haven't tested that yet.

Is there a reason why it's not in the same folder as the other .cmake files?

DrTimothyAldenDavis:
oops!  That's an oversight. It should be inside the SuiteSparse_config/cmake_modules subfolder.  I'll move it there now and post a 7.2.0.beta2.

PR #382, opened 2023-09-06
Don't generate .pc files for MSVC compatible tool chains.

mmuetzel:
The flags in the generated pkg-config files are incompatible with MSVC-compatible tool chains.
While it is very uncommon for MSVC-compatible tool chains to use .pc files at all, it would be better to not generate these files for those tool chains at all.

The proposed changes skip generating the .pc files when using a MSVC-compatible tool chain.

It would be nice if this could still be included in SuiteSparse 7.2.0.

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -196,28 +196,30 @@ install ( FILES
 # create pkg-config file
 #-------------------------------------------------------------------------------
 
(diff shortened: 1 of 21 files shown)


DrTimothyAldenDavis:
Yes - that would be a good idea if they're not needed for MSVC.  Thanks.  We can add this to SuiteSparse v7.2.0.

PR #383, opened 2023-09-06
SPQR: Overhaul CMake build rules.

mmuetzel:
Use targets more consistently instead of variables. 
Set properties on the target instead of globally where possible. That might be necessary anyway when adding a root CMakeLists.txt file.
No need for a CUDA compiler for the sources of SPQR. 
There are no OpenMP accelerated expressions in the sources. So, don't depend on OpenMP. I tried to grep OpenMP expressions using with the pattern `pragma.*omp` in the entire source tree of SuiteSparse. But there were only hits in the sources for GraphBLAS and CHOLMOD. I don't see how those sources could be built into the SPQR libraries.
Don't link transient targets of CHOLMOD (AMD and COLAMD) explicitly. 
Link static libraries to static libraries if possible.

Additionally, I was a bit surprised by the fact that SPQR is including files from GPUQREngine and SuiteSparse_GPURuntime that aren't installed by those subprojects.
Should those subprojects install those files? Are libraries linking against them supposed to be able to include them?
Alternatively, would it make sense to subsume those two libraries in SPQR? IIUC, only SPQR is using them anyway (if using CUDA). And the headers that are currently installed don't declare any symbols. The latter fact might mean that users of SuiteSparse aren't supposed to use functions from those libraries directly. Is that correct?
SPQR could still build them as separate libraries. There just wouldn't be stand-alone build rules. (Similarly to SPQR_CUDA.)

(This PR doesn't need to delay the release of SuiteSparse 7.2.0 imho.)

### DIFF
--- a/SPQR/CMakeLists.txt
+++ b/SPQR/CMakeLists.txt
@@ -46,6 +46,14 @@ project ( spqr
 # find library dependencies
 #-------------------------------------------------------------------------------
 
+option ( NOPENMP "ON (default): do not use OpenMP.  OFF: use OpenMP" ON )
+if ( NOPENMP )
+    # OpenMP has been disabled
+    set ( OPENMP_FOUND false )
+else ( )
+    find_package ( OpenMP )
+endif ( )
+
 find_package ( SuiteSparse_config 7.2.0
     PATHS ${CMAKE_SOURCE_DIR}/../SuiteSparse_config/build NO_DEFAULT_PATH )
 if ( NOT TARGET SuiteSparse::SuiteSparseConfig )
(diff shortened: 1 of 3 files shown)


DrTimothyAldenDavis:
SPQR might be using OpenMP directly in the future (I want to replace TBB with OpenMP, in particular).

The GPUQREngine and SuiteSparse_GPURuntime packages aren't inside SPQR at the moment, as you pointed out.  The SuiteSparse_GPURuntime as general a packages for all other SuiteSparse packages.  GPUQREngine is specific to SPQR.

User's don't need to include header files from those 2 packages.

This was setup by a student working with me on the SPQR + GPU project and it's mostly a historical artifact now.

mmuetzel:
Thanks for clarifying.

I added the switch for enabling/disabling OpenMP back to the CMake rules. But I defaulted to disabling OpenMP. That avoids over-linking. It should be straight forward to switch the default once OpenMP is actually used.
Instead of handling variables or changing (global) CMake settings, the OpenMP targets are used.
Since most of SPQR is C++ code, I figured it would be most reasonable to use the C++ OpenMP target. If OpenMP will actually be added in *C* code, it should be easy to change to the C OpenMP target.

With respect to potentially moving the GPUQREngine subproject to SPQR, I'd prefer to keep that change separate. (But I'd still like to base that potential change on top of this PR. So, I'll wait until this PR is merged before looking into that.)

DrTimothyAldenDavis:
Sounds great.  When I add OpenMP to SPQR, it would be only to C++ code.  

I can merge this in for v7.2.0.beta, right?

mmuetzel:
> I can merge this in for v7.2.0.beta, right?

I think it should be save enough to be merged.
But it doesn't fix any obvious bugs. So, it's up to you if you'd like to include it.

DrTimothyAldenDavis:
I'll merge it in now. It shouldn't conflict with my bug hunting for the octave + suitesparse v7.2.0 issue.

PR #387, opened 2023-09-08
CI: Add runner without OpenMP to MSVC job.

mmuetzel:
This adds a runner that builds without OpenMP. I chose the MSVC runner to not blow up the other matrices by another dimension. (And the macOS runners hosted by GitHub seem to be slower than the other hosted runners.)

This is mostly to demonstrate a build error. But it might make sense to add it anyway to make sure it doesn't regress once this error is fixed.


DrTimothyAldenDavis:
Yes, this would be a good idea to add.

PR #390, opened 2023-09-11
Fix pkg-config files on platforms without import libraries.

mmuetzel:
Stripping the file extension from the library names doesn't work correctly on platforms where `CMAKE_IMPORT_LIBRARY_SUFFIX` is empty. That is all but Windows.

This PR fixes that issue by checking for the file extension `CMAKE_IMPORT_LIBRARY_SUFFIX` only on Windows.

It would be nice if this change could still make it into 7.2.0.


DrTimothyAldenDavis:
Sure -- I can add this to SuiteSparse v7.2.0.  It's still a beta release, so I'll just make a v7.2.0.beta4.

PR #393, opened 2023-09-12
Don't link to CHOLMOD_CUDA explicitly

### LINKED ISSUE #392: pkgconfig files still require CUDA libraries
**Describe the bug**
Building Fedora without CUDA, we are still ending up with CUDA dependencies in some pkgconfig files:
```
./usr/lib64/pkgconfig/KLU_CHOLMOD.pc:Requires.private: KLU BTF CHOLMOD CHOLMOD_CUDA
./usr/lib64/pkgconfig/UMFPACK.pc:Requires.private: SuiteSparse_config AMD  CHOLMOD CHOLMOD_CUDA
```
In Fedora this generates an unsastifiable dependency on `pkgconfig(CHOMOD_CUDA)`.

**Expected behavior**
No CHOLMOD_CUDA entries in the .pc files.

mmuetzel:
CHOLMOD_CUDA is automatically pulled in either by the imported CMake target `SuiteSparse::CHOLMOD` or by the pkg-config file `CHOLMOD.pc` if it is needed.
There is no need to link to it explicitly when CHOLMOD is already linked.

This fixes #392.

### DIFF
--- a/KLU/CMakeLists.txt
+++ b/KLU/CMakeLists.txt
@@ -86,14 +86,6 @@ if ( NOT NCHOLMOD )
         find_package ( CHOLMOD 4.2.0 )
     endif ( )
 
-    if ( SUITESPARSE_CUDA )
-        find_package ( CHOLMOD_CUDA 4.2.0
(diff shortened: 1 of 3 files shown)


DrTimothyAldenDavis:
Thanks!

PR #394, opened 2023-09-12
CI: Keep ccache for compilation with and without OpenMP separate

mmuetzel:
It looks like the MSVC runners with and without OpenMP poison their caches.
Keep them separate (similar to the ones with and without CUDA on the Ubuntu runners).


DrTimothyAldenDavis:
I'll do a PR from dev2 to dev, so this gets included in the default branch.

PR #395, opened 2023-09-12
CHOLMOD: Add CHOLMOD_CUDA to dependencies in .pc file if needed

mmuetzel:
Oops.
The CHOLMOD.pc is actually missing the dependency to CHOLMOD_CUDA.
This PR should fix that.

It also removes a line that tries to display the value of a variable that no longer(?) exists.


mmuetzel:
Wait a second. An optional `-DSUITESPARSE_CUDA` might also be missing in that file...

mmuetzel:
It would be nice if this could still be included in 7.2.0.

DrTimothyAldenDavis:
Sure thing.  No rush to SuiteSparse v7.2.0.  It can stay in beta for a while.

Issue #396, opened 2023-09-12
GraphBLAS no longer compiles with CUDA enabled


DrTimothyAldenDavis:
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/6c966bd5112dedab5a7d5799166f7feb503b8483/GraphBLAS/CMakeLists.txt#L27

I'm getting back to the CUDA kernels for GraphBLAS.  They are not enabled in production.

If I enable CUDA in the GraphBLAS/CMakeLists.txt file, I get a build error.  The CUDA library gets configured but the top-level CMakeLists.txt files with an error (target "graphblascuda" not found).

This worked in GraphBLAS v8.0.2 (the cmake would build but there was a compilation error later on because of other issues in how complex types are handled, but that's another issue).  The build now fails earlier; see below.

@mmuetzel :  this is failing because of the recent updates to the GraphBLAS/CMakeLists.txt file, with the new `export (EXPORT GraphBLASTargets ...)` and `install ( EXPORT GraphBLASTargets ...)` commands.  Is there something missing there when CUDA is enabled?

```
hypersparse $ make
( cd build && cmake  .. && cmake --build . --config Release -j8 )
-- Building SuiteSparse:GraphBLAS version: v8.2.0, date: Sept 8, 2023
-- GraphBLAS C API: v2.0, date: Nov 15, 2021
-- Source:        /home/faculty/d/davis/dev2/SuiteSparse/GraphBLAS 
-- Build:         /home/faculty/d/davis/dev2/SuiteSparse/GraphBLAS/build 
-- The C compiler identification is GNU 9.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Install lib:     lib
-- Install include: include
-- Install bin:     bin
-- Install rpath:   
-- Build   rpath:   /home/faculty/d/davis/dev2/SuiteSparse/GraphBLAS/build
-- Build type:    Release 
-- Looking for a Fortran compiler
-- Looking for a Fortran compiler - /usr/bin/f95
-- The Fortran compiler identification is GNU 9.4.0
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Check for working Fortran compiler: /usr/bin/f95 - skipped
-- Fortran: /usr/bin/f95
-- Looking for a CUDA compiler
[...]

PR #397, opened 2023-09-12
GraphBLAS: Export CMake target for GraphBLAS_CUDA.

mmuetzel:
This PR renames the CMake target `graphblascuda` to `GraphBLAS_CUDA` and exports it as `SuiteSparse::GraphBLAS_CUDA`.

There was no `FindGraphBLAS_CUDA`. So, there might not actually be a reason to define variables like `GRAPHBLAS_CUDA_INCLUDE_DIR` when importing the target. But that way, it is pretty similar to the other targets.

I haven't tested yet. And iiuc, the CI won't test it either because CUDA is hard disabled in the build rules for GraphBLAS.

Might be fixing #396.

### DIFF
--- a/GraphBLAS/CMakeLists.txt
+++ b/GraphBLAS/CMakeLists.txt
@@ -63,7 +63,7 @@ if ( SUITESPARSE_CUDA )
     # with CUDA and RMM
     set ( CMAKE_C_FLAGS  "${CMAKE_C_FLAGS} -DSUITESPARSE_CUDA" )
     add_subdirectory ( CUDA )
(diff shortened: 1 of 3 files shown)


DrTimothyAldenDavis:
That did the trick.  CUDA for GraphBLAS is now (almost) compiling with this update.  To be precise, it works the same as GraphBLAS in SuiteSparse v7.1.0:  the cmake works but there's an unrelated error when trying to compile complex-typed kernels in CUDA.  That was already known.  This fixes the cmake problem -- thanks!

PR #399, opened 2023-09-13
Automatically look for CUDA targets when importing GraphBLAS

mmuetzel:
This is similar to what is done for CHOLMOD_CUDA.

Currently untested. (Still not set up to test anything CUDA locally.)


DrTimothyAldenDavis:
Thanks!  I'll give it a try.

PR #400, opened 2023-09-13
updates to build system (part 2; cuda)

DrTimothyAldenDavis:


mmuetzel:
`find_package ( ... REQUIRED )` inside a Config.cmake file doesn't do the right thing if the target is imported without the `REQUIRED` flag in a downstream project.
But we are currently using this idiom, e.g., to import the CUDAToolkit (and other targets that should be "transparent" to a user).

I've an idea how to improve that. But it's getting late here. I'll try to open a PR tomorrow.
Could you please wait another day before tagging the next beta?

DrTimothyAldenDavis:
Sure -- I'm in no rush to tag a new beta release.

mmuetzel:
I opened #402 for that change.

DrTimothyAldenDavis:
OK.  I'll merge both in -- thanks.

PR #402, opened 2023-09-14
CMake: Don't impose REQUIRED flag when importing targets

mmuetzel:
Use `find_dependency` instead of `find_package` to respect the options the user chose to find these targets (e.g., with or without `REQUIRED` flag).

This might look more complicated than it needs to be. But I'm planning to extend that at some point in the future. E.g., importing the CHOLMOD target could also import the CAMD target in case CHOLMOD was built with it. Having that structure set up now will hopefully make it more easy to build on top of that change.

### DIFF
--- a/CHOLMOD/Config/CHOLMODConfig.cmake.in
+++ b/CHOLMOD/Config/CHOLMODConfig.cmake.in
@@ -34,22 +34,37 @@ set ( CHOLMOD_VERSION_MINOR @CHOLMOD_VERSION_MINOR@ )
 set ( CHOLMOD_VERSION_PATCH @CHOLMOD_VERSION_SUB@ )
 set ( CHOLMOD_VERSION "@CHOLMOD_VERSION_MAJOR@.@CHOLMOD_VERSION_MINOR@.@CHOLMOD_VERSION_SUB@" )
 
+# Check for dependent targets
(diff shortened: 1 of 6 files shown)


DrTimothyAldenDavis:
Looks fine.  I'll merge it in -- thanks.

Shall I post another beta release of SuiteSparse 7.2.0?  I'd like to finalize that release soon, if things are ready to go.

mmuetzel:
I'm currently not aware of other bugs in the CMake build system. That doesn't mean there are none though.

I still have a couple of changes in the pipeline. But those don't need to block a release.

I'll continue testing different configurations when I come around to it. But most things seem to be working now. And in case something breaks horribly (I don't expect that currently), there can probably be a 7.2.1 to fix that.

DrTimothyAldenDavis:
Great!  Yes, things seem to be working well now.  I'll go ahead a post a formal 7.2.0 release, either today or tomorrow.

mmuetzel:
> I'm currently not aware of other bugs in the CMake build system. That doesn't mean there are none though.

That didn't age well... See #406.

DrTimothyAldenDavis:
So it goes :-)

I think my record for someone catching a bug after releasing a stable package (non-beta) is maybe just a few hours.  It was a GraphBLAS bug if I recall.

PR #404, opened 2023-09-15
CHOLMOD: Automatically check for dependencies when importing target

mmuetzel:
Users don't really need to know that CHOLMOD depends on AMD and COLAMD (and likely also on CAMD and CCOLAMD).

With the proposed changes, it should be enough for a user to just import the CHOLMOD target and link their own CMake target to it. The transitive dependencies should be resolved automatically  by CMake.

That is done for the SuiteSparse libraries that depend on CHOLMOD.

### DIFF
--- a/CHOLMOD/Config/CHOLMODConfig.cmake.in
+++ b/CHOLMOD/Config/CHOLMODConfig.cmake.in
@@ -61,34 +61,46 @@ endif ( )
 
 # Look for SuiteSparse_config, COLAMD, and AMD targets
 if ( @SUITESPARSE_IN_BUILD_TREE@ )
-    # First check in a common build tree
-    find_dependency ( SuiteSparse_config @SUITESPARSE_CONFIG_VERSION_MAJOR@.@SUITESPARSE_CONFIG_VERSION_MINOR@
-        PATHS ${CMAKE_SOURCE_DIR}/../SuiteSparse_config/build NO_DEFAULT_PATH )
-    # Then, check in the currently active CMAKE_MODULE_PATH
-    if ( NOT SuiteSparse_config_FOUND )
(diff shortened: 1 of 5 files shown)


DrTimothyAldenDavis:
Good point.  Can we save this for SuiteSparse 7.2.1 though?  I'd like to post a stable v7.2.0 soon, since I'll be out of commission for a while starting next week (surgery on Sept 20th...).

mmuetzel:
> Can we save this for SuiteSparse 7.2.1 though?

This can wait for sure. 
Imho, it's just a change for "comfort". Not really required given that it doesn't fix a regression or error...

PS: Best wishes for your surgery next week.

DrTimothyAldenDavis:
I'll post a 7.2.1 beta soon (perhaps this evening or tomorrow), with the fixes for amd_internal.h (and the related include files used by GPUQREngine), and your fix to Mongoose for SuiteSparse_config, so that those linux distro builds can get fixed as soon as possible.

This change can wait until a v7.2.2 release, right?

mmuetzel:
> This change can wait until a v7.2.2 release, right?

I agree. This can wait until later. (And it still needs a couple of follow-ups to entirely fix #409. I'd like to base those follow-ups on the already open PRs though.)

mmuetzel:
Now that 7.2.1 is released (thank you again for the new version), could this PR be merged? If that is ok, could that be followed up by merging #411, #412, and #413?
My other PRs will likely need rebasing after that. I'm happy to do that as a follow-up.

DrTimothyAldenDavis:
Yes -- I'll do those this week.  I first need to post a 2-line bug fix in CHOLMOD, for v7.2.2.  I just found it this weekend.  Not yet posted to dev2.

DrTimothyAldenDavis:
Now posted to dev2: https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/c0b141a984939884df4067df9994833c93547d22

DrTimothyAldenDavis:
Merging these into dev2 now.

PR #407, opened 2023-09-17
Mongoose: Link executables to SuiteSparse_Config that depend on it

### LINKED ISSUE #406: Mangoose doesn't link with DNSTATIC=1
**Describe the bug**
Mangoose fails to link when building suitesparse with DNSTATIC=1
```
/usr/bin/ld: libmongoose.so.3.2.0: undefined reference to symbol 'SuiteSparse_time'
/usr/bin/ld: /build/suitesparse/src/SuiteSparse-7.2.0/SuiteSparse_config/build/libsuitesparseconfig.so.7: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

**To Reproduce**
```
CMAKE_OPTIONS=" -DNSTATIC=ON" \
  make
```

**Expected behavior**
Builds.
Changing `PRIVATE` to `PUBLIC` in https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/dev/Mongoose/CMakeLists.txt#L212 fixes it.

 - OS: Linux x86_64
 - compiler gcc 13.2.1
 - BLAS and LAPACK library, if applicable
 - Version 7.2.0

**Additional context**

mmuetzel:
Some tests or demos in the Mongoose project use functions from SuiteSparse_Config.
When linking statically, that dependency happens to be pulled in via Mongoose_static. But when linking dynamically, that dependency doesn't happen to be pulled in "by chance".

Link target for executables that are using functions from SuiteSparse_Config explicitly to the respective target. (The build rules in that project seem to prefer static over shared linking. So, do the same here.)

Fixes #406.

### DIFF
--- a/Mongoose/CMakeLists.txt
+++ b/Mongoose/CMakeLists.txt
@@ -247,6 +247,11 @@ if ( NSTATIC )
 else ( )
     target_link_libraries ( mongoose_exe Mongoose_static )
 endif ( )
+if ( TARGET SuiteSparse::SuiteSparseConfig_static )
+    target_link_libraries ( mongoose_exe SuiteSparse::SuiteSparseConfig_static )
+else ( )
+    target_link_libraries ( mongoose_exe SuiteSparse::SuiteSparseConfig )
+endif ( )
 
 # Build the Demo executable
 add_executable ( demo_exe ${DEMO_FILES} )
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Thanks for the fix!

Issue #409, opened 2023-09-17
Missing dependencies between cmake config files


antonio-rojas:
**Describe the bug**
Using this `CMakeLists.txt`:
```
find_package(CXSparse)

add_library(foo)
target_sources(foo PRIVATE foo.cpp)
target_link_libraries(foo SuiteSparse::CXSparse)
```
cmake fails with
```
CMake Error at CMakeLists.txt:5 (target_link_libraries):
  Error evaluating generator expression:

    $<TARGET_PROPERTY:SuiteSparse::SuiteSparseConfig,INTERFACE_INCLUDE_DIRECTORIES>

  Target "SuiteSparse::SuiteSparseConfig" not found.
```
Looks like CXSparseConfig.cmake is missing a `find_dependency(SuiteSparse_config)` line. There may be other instances of the same issue in other libraries.

mmuetzel:
That is (still) expected behavior and no regression from previous versions of SuiteSparse.
You (still) need to manually use `find_package` for *all* dependent SuiteSparse libraries before being able to link to them.

That is currently being worked on. See, e.g., #404.

antonio-rojas:
Closing as duplicate

mmuetzel:
No need to close this report. That's still an issue that should be fixed imho.
I just wanted to note that it's not a regression nor a bug for which a fix needs to be rushed.

antonio-rojas:
Oh, I missed it was a PR for another similar issue.

mmuetzel:
I opened a couple of additional PRs that should be fixing this for a few more libraries.
That leaves KLU, SPQR and UMFPACK for which I'd like to wait for #404 to be merged.
Additionally, I didn't touch GPUQREngine which might be integrated into SPQR (#385).

mmuetzel:
I opened #443 and #444 for the dependencies when importing the targets for KLU and UMFPACK. (And #445 to add parts missing from previous PRs.)

That only leaves SPQR. I'll wait for #385 to be merged before changing that to avoid potential merge conflicts.

mmuetzel:
Afaict, this is fixed on the `dev2` branch now (apart from the two new projects that haven't been changed to create CMake import targets yet).

mmuetzel:
Should be working also for the ParU and LAGraph targets in the dev2 branch now.

Should this be closed now? Or wait until this makes it to a release?

DrTimothyAldenDavis:
[...]

PR #413, opened 2023-09-17
SPEX: Automatically check for dependencies when importing target

mmuetzel:
Similar to #404.

Additionally, simplify build rules for example by no longer importing the modules for GMP and MPFR explicitly.

Fixes part of #409.


DrTimothyAldenDavis:
looks great --  thanks again for all your help!

PR #419, opened 2023-09-19
Set preprocessor macro SUITESPARSE_CUDA on a per-target level

mmuetzel:
Instead of setting the preprocessor definition `SUITESPARSE_CUDA` globally, set it on a per-target level.

This avoids some issues when trying to build everything with a root CMakeLists.txt (in particular GraphBLAS which currently needs to deviate from the global setting).


mmuetzel:
Could this PR be merged? That would allow to get #415 in a more or less ready state to be considered for merging...

DrTimothyAldenDavis:
Merging it now.  Thanks!

PR #422, opened 2023-09-19
CI: Make sure that "with" doesn't match "without" in ccache key

mmuetzel:
I was wondering why some runners took exceptionally long compared to others sometimes.
Turns out that is because "with" can match "without" in the restore key for the GitHub cache that's used for ccache. (Didn't think of that 😖)

Make sure that doesn't happen by adding ":" to the affected restore keys.

This is a low-impact change that shouldn't affect the release process.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -86,7 +86,7 @@ jobs:
           # Prefer caches from the same branch. Fall back to caches from the dev branch.
           restore-keys: |
             ccache:ubuntu:${{ matrix.compiler }}:${{ matrix.cuda }}:${{ github.ref }}
-            ccache:ubuntu:${{ matrix.compiler }}:${{ matrix.cuda }}
+            ccache:ubuntu:${{ matrix.compiler }}:${{ matrix.cuda }}:
 
       - name: create empty libraries
         # This is to work around a bug in nvlink.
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Ah, yes, I was wonder about that.  I'll add this to dev2 and dev right away.

DrTimothyAldenDavis:
I've run out of run time that GitHub provides for CI.  It resets Oct 15th, unless I pay for more time.  This change will help reduce the time it takes to do the CI.  Too late for this month's cycle, unfortunately.

mmuetzel:
> I've run out of run time that GitHub provides for CI. It resets Oct 15th, unless I pay for more time.

Is there a limit for the maximum runtime for open source projects? I didn't run into that before. 
I'd understand their documentation that hosted runners are free to use for open source projects:
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

Could you please show a screenshot of that error message?

DrTimothyAldenDavis:
Oh, I must have hit that limit on an unrelated private repo just now.

No worries then.  The private repo doesn't need CI.

DrTimothyAldenDavis:
I think the run time used for open source repos must be free, but the time it uses for that is subtracted from time available for non-open repos.  So this update will help that.

However, I don't really need CI on this private repo anyway.

mmuetzel:
Sorry if I'm stupid. Which private repo?

Issue #425, opened 2023-09-22
Need ability to build and package multiple configurations of suitesparse


opoplawski:
**Is your feature request related to a problem? Please describe.**
Fedora currently builds 3 versions of SuiteSparse:
* 32-bit INT - linked with flexiblas
* 64-bit INT - linked with flexiblas
* 64-bit INT with 64_ symbol suffice linked with openblas (since flexiblas does not provide a 64_ symbol suffixed version)

I'm looking to update the Fedora package to the latest version and use the cmake build system.  But I'm running into a number of difficulties:

* SuiteSparse_config.h is different for the 32-bit and 64-bit blas:
```
--- no64/usr/include/suitesparse/SuiteSparse_config.h   2023-09-05 18:00:00.000000000 -0600
+++ empty/usr/include/suitesparse/SuiteSparse_config.h  2023-09-05 18:00:00.000000000 -0600
@@ -507,7 +507,7 @@
 #else
 
     // let CMake determine the size of the integer in the Fortran BLAS
-    #define SUITESPARSE_BLAS_INT int32_t
+    #define SUITESPARSE_BLAS_INT int64_t
 
 #endif
```
Not terribly surprising.  I think this can be handled by `-DCMAKE_INSTALL_INCLUDEDIR=/usr/include/suitesparse64`

* I am using `-DCMAKE_RELEASE${build}_POSTFIX=$build` with build set to "", "64", and "64_" to get the shared libraries to have the needed suffixes.
* The cmake files can be mostly differentiated by specifying different `CMAKE_BUILD_TYPE`s - `Release`, `Release64`, and `Release64_`.  This results in different `*Targets-Release{,64,64_}.cmake` files.  However, because of the difference between flexiblas and openblas linkage described above there are a couple of differences in the main cmake Targes files:
```
--- no64/usr/lib64/cmake/CHOLMOD/CHOLMODTargets.cmake   2023-09-05 18:00:00.000000000 -0600
+++ empty/usr/lib64/cmake/CHOLMOD/CHOLMODTargets.cmake  2023-09-05 18:00:00.000000000 -0600
@@ -76,7 +76,7 @@
 
 set_target_properties(SuiteSparse::CHOLMOD_static PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES "/usr/include/suitesparse"
[...]

PR #426, opened 2023-09-22
Allow using CMAKE_BUILD_TYPE and CMAKE_${CMAKE_BUILD_TYPE}_POSTFIX to…

opoplawski:
… add a postfix to pkgconfig files

This partially addresses #425 

See also https://discourse.cmake.org/t/how-to-evaluate-config-postfix/8984 for some discussion.


DrTimothyAldenDavis:
Thanks -- I'm running tests now.  I'm a bit behind (recovering from surgery 2 weeks ago, and I've been away from the keyboard for that whole time; only now slowly getting up to speed).

mmuetzel:
I don't think it is possible to "merge" different `*Target.cmake` files into one. So, I don't think that this change can conceptionally resolve the issue described in #425.
But please let me know if I'm missing something.

mmuetzel:
This is probably superseded by the different approach applied in #468.
Should this PR be closed?

DrTimothyAldenDavis:
Yes, I'll close this now.  Thanks for all the suggestions; it's been very helpful.

Issue #428, opened 2023-10-04
Cross compilation broke in 7.2


ViralBShah:
With the SuiteSparse 7.2 release, it seems that some compile time tests are being carried out, which were not an issue earlier. If possible, it would be great to avoid `try_run()` in Cmake so that SuiteSparse can be cross-compiled easily. In Julia's BinaryBuilder, we cross compile binaries for a large number of architectures on linux.

https://github.com/JuliaPackaging/Yggdrasil/pull/7479
```
2023-10-03 21:06:13 UTC | [21:06:14] CMake Error: try_run() invoked in cross-compiling mode, please set the following cache variables appropriately:
-- | --
  | 2023-10-03 21:06:14 UTC | [21:06:14]    HAVE_KEYWORD__THREAD_EXITCODE (advanced)
  | 2023-10-03 21:06:14 UTC | [21:06:14]    HAVE_KEYWORD__THREAD_EXITCODE__TRYRUN_OUTPUT (advanced)
  | 2023-10-03 21:06:14 UTC | [21:06:14] For details see /workspace/srcdir/SuiteSparse/CHOLMOD/build/TryRunResults.cmake
```

imciner2:
Caused by the tests for the thread attributes here: https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/75cb8b4be44227880388454dba022bc7442de61e/SuiteSparse_config/cmake_modules/SuiteSparse__thread.cmake#L46 The CMake command `check_c_source_runs` will both compile and run the code, so that is what is failing here.

There is an alternate `check_c_source_compiles()` (https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html) CMake macro that will just check to see if the code can compile, and not run it (so that only tests the compiler), and based on the short snippets being used in the test, I think it would be safe to do the compile-only test instead of the full run test.

DrTimothyAldenDavis:
I have several of the check_c_source_runs:  three in the `SuiteSparse__thread.cmake` file, and another here:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/75cb8b4be44227880388454dba022bc7442de61e/GraphBLAS/cmake_modules/SuiteSparse_getenv.cmake#L26

That one for getenv does require "runs" not "compiles" since I must test not only the existence of getenv, but that it also recognizes the "HOME" string.  I cannot just test that the code compiles.  This may be a problem for cross-compilation, however.

[...]

PR #429, opened 2023-10-05
Fix the ordering of the __thread and static specifiers

imciner2:
GCC requires that the `__thread` keyword appear immediately after the storage class specifier.

ref: https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html, which says:

> The __thread specifier may be used alone, with the extern or static specifiers, but with no other storage class specifier. When used with extern or static, __thread must appear immediately after the other storage class specifier.

### DIFF
--- a/CHOLMOD/SuiteSparse_metis/GKlib/random.c
+++ b/CHOLMOD/SuiteSparse_metis/GKlib/random.c
@@ -71,8 +71,8 @@ GK_MKRANDOM(gk_z,   size_t, ssize_t)
 #elif defined ( HAVE_KEYWORD__THREAD )
 
     // gcc and many other compilers support the __thread keyword
-    __thread static uint64_t mt[NN];
-    __thread static int mti=NN+1;
+    static __thread uint64_t mt[NN];
+    static __thread int mti=NN+1;
 
 #elif defined ( HAVE_KEYWORD__DECLSPEC_THREAD )
 


mmuetzel:
LGTM. 👍

DrTimothyAldenDavis:
Looks good.  I will go ahead and merge it once the tests pass.

DrTimothyAldenDavis:
Related to https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/428

mmuetzel:
I'm not sure if it's worth waiting for the CI. Afaict, there is no runner that would check that code path.
That code path is only used when building without OpenMP. The only runner that does that is using MSVC. MSVC doesn't support the __thread keyword afaict. So, that code path isn't used.

Maybe, we should add more runners that build without OpenMP (including a GCC runner)? That might have caught that issue earlier...

DrTimothyAldenDavis:
I went ahead and merged it before the tests finished.

DrTimothyAldenDavis:
Yes, gcc on ubuntu without OpenMP would have caught this error.

Interestingly, clang has no problem with this.  The following code compiles on my Mac with clang but not with gcc 13.  gcc-13 is ok with the definition of x, but gives an error on the line defining y.

```
#include <stdio.h>

    static __thread int x = 1 ;   // the correct order
    __thread static int y = 1 ;   // causes gcc to fail; clang is fine

    int main (void)
    {
        x = 0 ;
        y = 3 ;
        printf ("hi %d %d\n", x, y) ;
        return (x) ;
    }

```

gcc reports:

```
gunk.c:5:5: error: '__thread' before 'static'
    5 |     __thread static int y = 1 ;
      |     ^~~~~~~~
```

imciner2:
> Interestingly, clang has no problem with this. The following code compiles on my Mac with clang but not with gcc 13

Yea, this is one of the GCC extensions to the C spec, so the difference in behavior is to be expected, unfortunately, even more so with Apple Clang because they take upstream Clang and make their own opaque changes to it.

PR #430, opened 2023-10-06
GraphBLAS: Remove test for which the result is inconsequential

mmuetzel:
~~Use the same cache directory as a fallback on Windows that is used in
`GB_jitifyer.c` also in `GB_jit_cache.cu`.~~

Also remove a test for which the result is inconsequential. (See https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/428#issuecomment-1751118631)


DrTimothyAldenDavis:
Looks great, except that the CUDA file shouldn't be changed.  That file needs more work, and isn't in production use.  It needs to be reworked so that it uses the GB_jit_cache determined by GraphBLAS/Source/GB_jitifyer.c.  I have lots of work to do on the CUDA folder, and that work is ongoing in the draft v9.0.0.  No need to fix it here for v8.2.1.

mmuetzel:
> No need to fix it here for v8.2.1.

But I'd guess there is also no harm in making that change. (That part might disappear when the other issue gets fixed. But it's not strictly wrong to change that file for the time being either.)

DrTimothyAldenDavis:
It gets complicated because I already have a v9.0.0 draft of GraphBLAS, and it complicates the version control from one to the other.  So I'd rather not change it at all.

mmuetzel:
> So I'd rather not change it at all.

I stripped that commit from the PR.

DrTimothyAldenDavis:
I'll merge this in once the tests pass, and then revert the change to the CUDA folder.

edit:  oh, never mind.  I just saw your comment above that you already removed it -- thanks!

PR #431, opened 2023-10-07
CI: Add Ubuntu runner using GCC without OpenMP

mmuetzel:
Instead of blowing up the build matrix with another dimension, I opted to add only one single runner that builds without OpenMP on Ubuntu. I opted to build with CUDA on that runner because I thought the code coverage would be larger that way.

Please, let me know if you think this should be done differently.

I guess that new runner would have caught the issue fixed in #429.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,7 +22,7 @@ jobs:
     # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
     runs-on: ubuntu-latest
 
-    name: ubuntu (${{ matrix.compiler }} ${{ matrix.cuda }} CUDA)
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Without CUDA would probably be better, because I would guess that any system that has CUDA must also have OpenMP.  So no-OpenMP-but-with-CUDA would be an unusual case.  It would be more typical to have (a) OpenMP and CUDA, (b) OpenMP but not CUDA, or (c) neither OpenMP nor CUDA.  We already test for (a) and (b) but not (c).

So using gcc on ubuntu without OpenMP and without CUDA would be a better test.  I'm not so worried about the test coverage.  It would be more important to test a more common case.

mmuetzel:
> So using gcc on ubuntu without OpenMP and without CUDA would be a better test.

Made that change.

PR #453, opened 2023-10-20
AMD: Don't force-disable Fortran on Windows

mmuetzel:
I don't know the original motivation why Fortran was deactivated for AMD on Windows unconditionally. But it is working for me now. 
So, remove the logic that deactivates Fortran for AMD on Windows.

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -35,11 +35,6 @@ include ( SuiteSparsePolicy )
 # define the project
 #-------------------------------------------------------------------------------
 
-if ( WIN32 )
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
I think a while back, Windows didn't have a Fortran compiler.  It could be renabled, and if someone has trouble on Windows they can set the flag back on.

mmuetzel:
Oops. `NFORTRAN` is only defined after `SuiteSparsePolicy` has been included...

I'll post a follow up to fix that soon.

Edit: Wrong PR. That comment was meant for #415.

PR #454, opened 2023-10-20
Alow to find libflame

thierry-FreeBSD:
SuiteSparse does not use the standard module FindLAPACK.cmake installed by cmake, but a special file SuiteSparseLAPACK.cmake, and when configured with Blis / FLAME, it forces the use of the generic lapack library.

This proposed PR let it use the libflame library.


mmuetzel:
Wouldn't the same change be needed for LAPACK_STATIC a couple of lines down?

While you are at it, could you also fix the error in the comment (liblagraph.a)? That should probably say "liblapack.a" (or "libflame.a" with your changes).

mmuetzel:
@DrTimothyAldenDavis: Is there a reason why the default [`FindLAPACK.cmake`](https://cmake.org/cmake/help/latest/module/FindLAPACK.html) isn't used for libflame? Does that cause issues?

DrTimothyAldenDavis:
Looks great -- thanks!

I had started with FindLAPACK but had a lot of trouble getting it to work, particularly in how it was managing 32-bit vs 64-bit versions.  By that, I don't mean the size of a pointer.  I mean the size of the integers that are passed to BLAS and LAPACK.

I prefer to use BLAS and LAPACK libraries that allow for the use of 64-bit integer parameters for n, m, lda, and other integer parameters.  That allows me to solve larger problems.  I also have to pick the right size because the BLAS and LAPACK will fail if a C code calls the Fortran codes with the wrong sizes.

Ideally, I would use the CBLAS interface ... but that only supports 32-bit integers.

I wasn't able to get SuiteSparseLAPACK.cmake to work with the BLIS / FLAME libraries, because all I got was the BLAS, not LAPACK.  But if this fixes it, I'm very happy to include this change.

Once the CI passes, I'll merge it in.  Thanks for the help!

DrTimothyAldenDavis:
I can fix that typo once I merge it in.

mmuetzel:
Should line 38 in that file also be changed? (I'd guess it should.)

thierry-FreeBSD:
> Should line 38 in that file also be changed? (I'd guess it should.)

It seems so - but I have not tested with static libraries.

DrTimothyAldenDavis:
@thierry-FreeBSD :  I've included your fix to SuiteSparseLAGRAPH.cmake in SuiteSparse 7.3.0.beta1.  Can you give it a try?

thierry-FreeBSD:
Yes, I can confirm, it works, with libblis and libflame from the BLIS Framework (FLAME).

libblis from AMD (AOCL-BLAS) can be used too, but I have not yet checked AOCL-LAPACK.

I'm investigating various BLAS implementations, and I shall propose another patch to support FlexiBLAS (not yet ready).

DrTimothyAldenDavis:
Great!  Thanks for checking it out.  I will go ahead and post 7.3.0 as a stable release shortly.

PR #455, opened 2023-10-20
ParU: First pass for CMake build rules

mmuetzel:
It is building now in most of the configurations of the CI. 
It's only still failing for MSVC with OpenMP. Maybe, we need some additional flags for that? But it's getting late here. And the current state might be enough to share.

This library also seems to include a header that isn't installed: `umf_internal.h` which in turn includes `amd_internal.h`. Should these be copied again? (I don't really like that workaround...)


DrTimothyAldenDavis:
Ah, yes, umf_internal.h again.  For now, the umf_internal.h and amd_internal.h files can be copied into the ParU/Include folder.

I will work on a more robust solution.

PR #459, opened 2023-10-23
LAGraph: Re-run failing tests with more verbose output and continue.

mmuetzel:
Re-run failing tests with `--rerun-failed --output-on-failure` to get more verbose output that might help in tracking down the issue on 32-bit platforms.

Also continue even if ctest fails to make it behave more similar to the other projects.

### DIFF
--- a/LAGraph/Makefile
+++ b/LAGraph/Makefile
@@ -63,7 +63,7 @@ debug:
 all: library
 
 test: library
-	( cd build && ctest . )
+	( cd build && ctest . || ctest . --rerun-failed --output-on-failure || true )
 
 # target used in CI
 demos: test


DrTimothyAldenDavis:
I see the bug.  LAGraph_Malloc has a test that tries to avoid calling malloc when the request is too large:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/fd163d9526ca240bb569699907ea04eb695a1846/LAGraph/src/utility/LAGraph_Malloc.c#L42-L48

but `GrB_INDEX_MAX` is always 2^60 on all platforms.  On a 32-bit platform, `size_t` is 4 bytes, however, and the test_Malloc script does this on line 39-40, which is supposed to trigger that test and return p = NULL.

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/fd163d9526ca240bb569699907ea04eb695a1846/LAGraph/src/test/test_Malloc.c#L39-L40

That input parameter, `GrB_INDEX_MAX+1`, gets typecasted from uint64_t to size_t, which truncates it, probably to the value 1.  So the call to LAGraph_Malloc succeeds, when it is supposed to fail.

I will need to look through all of LAGraph to make sure there's no other broken typecasting like this with `size_t`.

PR #460, opened 2023-10-23
CI: Build (and test) GraphBLAS and LAGraph also on x86

mmuetzel:
To reduce build time, GraphBLAS is built in COMPACT mode. But even with the COMPACT mode, building GraphBLAS on emulated hardware takes too long. So, it's only built on x86 (where it finishes in a reasonable time).

It looks like the test errors with LAGraph aren't limited to 32-bit Windows. They are also failing for 32-bit Linux. (For some reason, two tests are failing on 32-bit Windows, and three tests are failing on 32-bit Linux.)

### DIFF
--- a/.github/workflows/build-arch-emu.yaml
+++ b/.github/workflows/build-arch-emu.yaml
@@ -34,7 +34,9 @@ jobs:
         arch: [x86, aarch64, armv7, ppc64le, s390x]
         include:
           - arch: x86
-            ccache-max: 42M
+            ccache-max: 80M
+            extra-build-libs: ":GraphBLAS:LAGraph"
+            extra-check-libs: ":GraphBLAS:LAGraph"
           - arch: aarch64
             ccache-max: 42M
           - arch: armv7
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Thanks -- I'll take a closer look.  I don't think we've tested LAGraph on 32-bit platforms yet.

DrTimothyAldenDavis:
ParU uses OpenMP tasking, which requires OpenMP 4.0, I think.  MSVC only supports OpenMP 2.0 if I remember correctly.

I was able to work around this issue when I encountered it in GraphBLAS, by avoid the use of OpenMP tasking entirely.  That's not possible with ParU -- it requires OpenMP tasking as a fundamental part of its methods.  So it will never work with MSVC with OpenMP.

@Aznaveh : ParU can be compiled without OpenMP, correct?

So I think the simplest solution would be to disable OpenMP, just for ParU, on MSVC.  We should add a compile-time test to ParU (in the code or in the cmake), so that it will not try to use any OpenMP at all if the OpenMP version is too low.

PR #465, opened 2023-10-23
LAGraph: Fix linking with MSVC

mmuetzel:
The variables for the `dllexport`/`dllimport` attributes must be defined on a per library basis. They must not be added when building a static library.
Additionally, don't mark global variables as `extern` where they are defined.
Function declarations are `extern` implicitly. Add `extern` keyword only at global *variable* declarations. Add `dllexport`/`dllimport` attributes to global variable definitions where they are missing.

Also don't use `WINDOWS_EXPORT_ALL_SYMBOLS` for this project. This essentially reverts https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/301279187f5ed15bc8344092fe70d3ff8b35a61f.


mmuetzel:
It looks like the `msvc (without OpenMP)` runner is green with these changes now.

DrTimothyAldenDavis:
Looks great!

Is the LAGRAPHX_PUBLIC needed, or could we just use LAGRAPH_PUBLIC in its place?

mmuetzel:
Each DLL needs its own set of preprocessor variables to be able to distinguish when the DLL is built or when it is being linked to.

Issue #462, opened 2023-10-23
ParU fails with OpenMP from MSVC


mmuetzel:
**Describe the bug**
When trying to build ParU with an MSVC toolchain, compilation fails with errors like the following:
```
  D:\a\SuiteSparse\SuiteSparse\ParU\Source\paru_analyze.cpp(1060): error C3001: 'taskloop': expected an OpenMP directive name
  D:\a\SuiteSparse\SuiteSparse\ParU\Source\paru_analyze.cpp(1068): error C3001: 'taskloop': expected an OpenMP directive name
  D:\a\SuiteSparse\SuiteSparse\ParU\Source\paru_analyze.cpp(1300): error C3001: 'taskloop': expected an OpenMP directive name
  D:\a\SuiteSparse\SuiteSparse\ParU\Source\paru_analyze.cpp(1314): error C3001: 'taskloop': expected an OpenMP directive name
  [3/84] Building CXX object CMakeFiles\paru.dir\Release\Source\paru_assemble.cpp.obj
  FAILED: CMakeFiles/paru.dir/Release/Source/paru_assemble.cpp.obj 
```
Afaict, the `taskloop` construct was introduced with OpenMP 4.5. MSVC supports only OpenMP 2.0.

**To Reproduce**
Try to build ParU with OpenMP using MSVC (see CI) or another toolchain that lacks support for OpenMP 4.5.

**Expected behavior**
ParU builds without error.

**Desktop (please complete the following information):**
 - OS: Windows
 - compiler MSVC 2022

DrTimothyAldenDavis:
Also tagging @Aznaveh.

The solution for this is to detect if OpenMP is 4.5 or later, and only use it for ParU for that version of OpenMP.  If OpenMP is an older version, like OpenMP 2.0 in Microsoft Visual Studio, then ParU needs to be compiled without OpenMP.

So that's a bug in our cmake -- we should be detecting this case in the cmake build system, and not use OpenMP for ParU if the version is too old.

The rest of SuiteSparse is fine with OpenMP 2.0 (CHOLMOD uses it in a few places,and GraphBLAS is fundamentally parallel).  I did use tasking in one place in GraphBLAS once, but I was able to change my algorithms to use the more basic OpenMP 2.0 methods.

The parallel algorithms in ParU can't be redesigned, however.  They are much more complex than my parallel algorithms in GraphBLAS and the will only work with OpenMP tasking.

[...]

PR #467, opened 2023-10-24
LAGraph: Adapt for stricter implementations of the OpenMP standard.

mmuetzel:
IIUC, the OpenMP standard requires that the iterator of parallel for loops is a signed integer that is declared outside the scope of the parallel loop body. Some implementations seem to have relaxed that requirement (allowing unsigned integers or initialization in the for loop). But others (MSVC) seem to follow a stricter interpretation of the standard.

Adapt the source to allow for that stricter interpretation.

Fixes the second issue raised in #462.


mmuetzel:
Looks like the CI is going to be green again with this change. 🎉

DrTimothyAldenDavis:
Thanks!  Yes, those "for (int k = .." loops don't work on Windows.  I'm not sure how those slipped through our CI on the LAGraph repo itself.  I knew about that but this LAGraph code was written by lots of people (I wrote the sort though so that's my bad).  I wasn't aware that OpenMP required the for loop variable to be a signed integer.  I'll have to double check my many parallel for loops in GraphBLAS (340 of them!); I tend to use signed integers but an unsigned might have slipped through.

PR #468, opened 2023-10-24
Support installing package configuration files at different location

mmuetzel:
This change allows the user to define a location for the CMake and pkg-config configuration files while installing the libraries, e.g., at a "default" location.

This will hopefully help for the use case described in #425.

@opoplawski: Would that work for your use-case?
You could, e.g., configure with 
```
cmake -DSUITESPARSE_INCLUDEDIR=include/suitesparse64 -DSUITESPARSE_PKGFILEDIR=lib/suitesparse64 -DCMAKE_RELEASE_POSTFIX=64 [other flags for that configuration] ..
```
That would allow to install the headers to a separate location like you already proposed in #425. Additionally, it would allow to install the CMake package configuration files and the .pc files in a location that differs from the "usual" one.
However, the libraries themselves would still be installed at the default location (with the SUFFIX that you were using). So, no RPATH or LD_PRELOAD shenanigans would be needed.

When building projects that rely on the "non-default" variants, users would need to set one of the following (depending on their choice of build system):
```
cmake -DCMAKE_PREFIX_PATH=/usr/lib/suitesparse64/cmake
PKG_CONFIG_PATH=/usr/lib/suitesparse64/pkgconfig
```

What do you think?

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -171,7 +171,7 @@ export ( EXPORT AMDTargets
 # install export target, config and version files for find_package
 install ( EXPORT AMDTargets
     NAMESPACE SuiteSparse::
-    DESTINATION ${SUITESPARSE_LIBDIR}/cmake/AMD )
+    DESTINATION ${SUITESPARSE_PKGFILEDIR}/cmake/AMD )
 
 # generate config file to be used in common build tree
 set ( SUITESPARSE_IN_BUILD_TREE ON )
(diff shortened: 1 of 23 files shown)


DrTimothyAldenDavis:
Looks great to me, but the pkgconfig setup is not something I know a lot about, so I'll defer to both @mmuetzel and @opoplawski on this question.

DrTimothyAldenDavis:
Shall I go ahead and merge it in, or do we want more comments first?

mmuetzel:
If I understand the use-case described by @opoplawski in #425 correctly, this change would probably help to have those three variants of SuiteSparse installed at the same time. Also building depending projects should be straight forward after this change (if I'm not missing something important).
Running projects that depend on any of those variants should be working out-of-the-box afaict.
So, from my point of view, this could be merged.
If something is missing, this could still be refined later (or reverted if it doesn't help at all).

DrTimothyAldenDavis:
OK -- merging it in, and if need be we can revise it later.

Issue #461, opened 2023-10-23
LAGraph tests fail on 32-bit platforms


mmuetzel:
Some of the ctests of LAGraph fail on 32-bit platforms (Windows or Linux).
E.g.: https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/6616841890/job/17971912574#step:8:5841
```
  95% tests passed, 3 tests failed out of 56
  
  Total Test time (real) = 141.18 sec
  
  The following tests FAILED:
  	  8 - ctest_ConnectedComponents (Failed)
  Errors while running CTest
  	 16 - ctest_Malloc (Failed)
  	 39 - ctest_BF (SEGFAULT)
```

**To Reproduce**
Build LAGraph for a 32-bit platform and run `ctest .` in the build directory.

**Expected behavior**
All tests should pass.

**Desktop (please complete the following information):**
 - OS: observed in the CI with Windows MINGW32, CLANG32 and Alpine Linux x86
 - compiler: `gcc.exe (Rev2, Built by MSYS2 project) 13.2.0` or `clang version 17.0.2 (Target: i686-w64-windows-gnu)` or `gcc (Alpine 12.2.1_git20220924-r10) 12.2.1 20220924`
 - all with OpenBLAS

DrTimothyAldenDavis:
I've fixed the test_Malloc.  I'm unsure about the test_BF and test_ConnectedComponents since those tests look to be passing now.  Is this an intermittent bug?

mmuetzel:
For easier reference: @DrTimothyAldenDavis is likely referring to this change:
https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/e6a5b01877490f043ecd2a3d81bcbe52f1a29f24

mmuetzel:
I rebased #415 to trigger the CI with your changes.

One test still segfaults on MINGW32:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/6623534748/job/17990771240?pr=415#step:9:7378
```
  98% tests passed, 1 tests failed out of 56
  
  Total Test time (real) = 146.46 sec
  
  The following tests FAILED:
  	 39 - ctest_BF (SEGFAULT)
  Errors while running CTest
```

The Alpine Linux x86 runner still fails the following tests:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/6623534747/job/17990770571#step:8:6136
```
  The following tests FAILED:
[...]

PR #469, opened 2023-10-25
LAGraph: Relax check for negative-weight cycle by floating point precision

mmuetzel:
Comparisons between floating point numbers can be "dangerous" at times because small errors could accumulate because of their limited precision.
If I understand the condition that checks for negative-weight cycles correctly, it shouldn't trigger if the compared numbers are equal.
This change allows them to be equal to double floating point precision.

This fixes part of #461.


DrTimothyAldenDavis:
Great!  Yes, that's a fine fix.

PR #473, opened 2023-10-29
LAGraph: Use `memmove` when condensing array `Tj`.

mmuetzel:
Use `memmove` (or if available `memmove_s`) instead of `for` loop.
That could be using a potentially more efficient implementation to do that job.

See also #461.

### DIFF
--- a/LAGraph/src/algorithm/LG_CC_FastSV6.c
+++ b/LAGraph/src/algorithm/LG_CC_FastSV6.c
@@ -47,6 +47,9 @@
 // G->A will then become a truly read-only object (assuming GrB_wait (G->A)
 // has been done first).
 
+#define __STDC_WANT_LIB_EXT1__ 1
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Thanks.  I commented out the memmove_s, since at this point in the code, I know that it won't copy anything outside the limits of the array, and the pointers for the source and destination cannot be NULL.  That would be caught earlier in the call to GxB_Matrix_unpack_CSR, which creates Tj as non-NULL or returns a failure code that is then caught by the GRB_TRY(...).

mmuetzel:
For later reference: @DrTimothyAldenDavis is likely referring to this change:
https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/26d4afb2e018d00b0f93ee5f9584fabc2f194b71

PR #475, opened 2023-10-29
ParU: Add CMake config and pkg-config files

mmuetzel:
Similar to the other subprojects of SuiteSparse, add CMake Config files that can be used to import CMake targets (instead of the CMake module file).
Additionally, add rules to create a pkg-config file for this library. Could you please check if the `Description` line in that file is appropriate?

I changed the CMakeLists.txt of the example to include ParU. This is currently only to see if the CMake module gets picked up correctly. The example doesn't actually call a function of the paru library yet.
@DrTimothyAldenDavis: It's probably much easier for you than for me to come up with a small self-contained example that could be added to `my.c` and `my_cxx.cc`. Would you be able to look into that?


DrTimothyAldenDavis:
Thanks!  Yes, I and/or @Aznaveh will take a look and add a call to ParU from the Example project.

PR #479, opened 2023-10-31
LDL: Avoid overlinking

mmuetzel:
The ldl library doesn't actually depend on libsuitesparseconfig or libamd. Only some of the demos do and some preprocessor definition from the SuiteSparse_config header are used in some compilation units.
Avoid overlinking that library to the libraries that it doesn't use.

Also avoid inclusion of a header inside an `extern "C"` block.


DrTimothyAldenDavis:
Good point.  Yes, ldl itself doesn't call AMD, just the demo program does.

PR #482, opened 2023-10-31
GraphBLAS: Avoid warnings about undefined `_OPENMP`

mmuetzel:
When building GraphBLAS without OpenMP, `gcc` emits a lot of preprocessor warnings about `_OPENMP` not being defined. The same happens for the JIT compiler.

Re-order some conditions in `GB_atomics.h` to omit these warnings.


GraphBLAS/Source/Template/GB_atomics.h
```diff
@@ -100,9 +100,10 @@
 // is no need for atomic reads/writes when compiling GraphBLAS on Windows
 // with MS Visual Studio.
 
-#if GBX86
+#if GBX86 || ! defined (_OPENMP) ||  (_OPENMP <= 201107 )
```
DrTimothyAldenDavis: That last term at the end of the expression should be (_OPENMP < 201107) I think.

GraphBLAS/Source/Template/GB_atomics.h
```diff
@@ -62,7 +62,12 @@
 // OpenMP 4.0 or later) with the "update" clause.  For MS Visual Studio, the
 // "update" clause is removed since it supports OpenMP 2.0.
 
-#if ( _OPENMP >= 201307 )
+#if ! defined (GB_ATOMIC_UPDATE) || (GB_ATOMIC_UPDATE < 199810)
```
DrTimothyAldenDavis: I think line 65 should be

#if ! defined ( _OPENMP) || (_OPENMP < 199810)

DrTimothyAldenDavis (commented):
This is a good idea, but there are 2 glitches below.

GraphBLAS/Source/Template/GB_atomics.h
mmuetzel: Oops. My bad.
Should be fixed now.

GraphBLAS/Source/Template/GB_atomics.h
mmuetzel: You're absolutely right. Probably mis-clicked when copying and pasted without checking...
Should be fixed now.

mmuetzel:
The compilation is more silent now. But the demos still spew out warnings like before:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/6709580738/job/18236014925#step:9:6102
```
  In file included from /home/runner/.SuiteSparse/GrB8.2.1/src/GB_Template.h:55,
                   from /home/runner/.SuiteSparse/GrB8.2.1/src/GB_jit_kernel.h:16,
                   from /home/runner/.SuiteSparse/GrB8.2.1/c/75/GB_jit__apply_unop__0000eb0ebe__mycx_cmplx_real.c:11:
  /home/runner/.SuiteSparse/GrB8.2.1/src/GB_atomics.h:65:7: warning: "_OPENMP" is not defined, evaluates to 0 [-Wundef]
     65 | #if ( _OPENMP >= 201307 )
        |       ^~~~~~~
  /home/runner/.SuiteSparse/GrB8.2.1/src/GB_atomics.h:70:9: warning: "_OPENMP" is not defined, evaluates to 0 [-Wundef]
     70 | #elif ( _OPENMP >= 201107 )
        |         ^~~~~~~
  /home/runner/.SuiteSparse/GrB8.2.1/src/GB_atomics.h:75:9: warning: "_OPENMP" is not defined, evaluates to 0 [-Wundef]
     75 | #elif ( _OPENMP >= 199810 )
        |         ^~~~~~~
[...]
```

Where does that come from? Does it need to be changed somewhere else, too?

DrTimothyAldenDavis:
If the source of GraphBLAS changes then a "make" is required inside the JITpackage folder. That is a compressed set of source files that gets added to the libgraphblas.so so the JIT can find its source at run time.  Then add that to the PR. 

The runner doesn't keep /home/username/.SuiteSparse/GrB8.2.1 I assume, or the LOCALAPPDATA/SuiteSparse/GrB8.2.1 folder I assume. But you would need to delete your copies of that folder. 

The end user doesn't need to do this because if a new version of GraphBLAS is installed, the old arc folder and old compiled jit kernels are abandoned (because the GrB version is in the name of the folder).

mmuetzel:
Thank you for the clear instructions. I wasn't aware of that.

DrTimothyAldenDavis:
Yes, it's an unusual mechanism.  The GraphBLAS JIT needs access to its own source code.  Rather than relying on source files outside of libgraphblas.so (which might not be available or perhaps stale), I incorporate compressed source code into libgraphblas.so itself.  When GraphBLAS starts, GrB_Init looks in ~/.SuiteSparse/GrB8.2.1 (or whatever version is being used) and ensures the src/* files are valid.   It can then create new JIT kernels in the c/* directories at run time, compile them, and load them in, on the fly.

The JIT kernels allow the end user to define new operators at run time, for me to use inside my kernels.  With the JIT, these kernels are just as fast as built-in kernels.

mmuetzel:
Would it make sense to make that part of the default build process? I.e., update the compressed JIT source whenever GraphBLAS is built?

DrTimothyAldenDavis:
Yes, that would make sense.  I wasn't sure how to go about doing that in a portable way.

It's rather like a very complex "configure_file (...)" process, but it needs to first compile the grb_jitpackage.c main program, which uses the GraphBLAS/zstd source code for the ZSTD compression package.

mmuetzel:
I guess that should be possible. I'll try to look into it after this PR is merged.

PR #483, opened 2023-10-31
Make example less dependent on layout of SuiteSparse source tree

mmuetzel:
This change should make the example more useful for downstream users. They likely aren't putting their sources in a common source tree with the SuiteSparse packages.


mmuetzel:
Waiting for #484 for a final change (remove the copy of the CMake modules in the example).

mmuetzel:
When you come around to it, could you please merge this PR. Atm, this is the last change needed to unlock #415 afaict. (But there might be more surprises...)

DrTimothyAldenDavis:
Thanks -- looks great.

PR #484, opened 2023-10-31
SPEX: Install CMake modules that are used in CMake Config file.

mmuetzel:
The CMake Config file of SPEX includes logic to find all its dependencies automatically when a user adds `find_package ( SPEX )` in their CMake project. That includes GMP and MPFR. However, the CMake modules needed to detect these two libraries haven't been installed.

Install them to the same location as the Config file to avoid errors when users include `find_package ( SPEX )` in their projects.


mmuetzel:
Could you please merge this PR? It would unblock a couple other changes.

DrTimothyAldenDavis:
Yes ... I've just been tied up (in all day meetings today through Friday).

mmuetzel:
Thank you. 👍

PR #486, opened 2023-10-31
BTF: Avoid overlinking

mmuetzel:
Pretty much the same as #479, but for BTF.


mmuetzel:
Not sure what happened to the macos runner. It looks it timed out after 6 hours. But the logs are partly missing...

DrTimothyAldenDavis:
I'll restart those jobs to see if they finish OK.

PR #487, opened 2023-10-31
ParU: Avoid overlinking

mmuetzel:
ParU itself doesn't depend on CHOLMOD. Only the demos do. Similar to #479.

Also fix a copy-paste error.

### DIFF
--- a/ParU/CMakeLists.txt
+++ b/ParU/CMakeLists.txt
@@ -219,15 +219,11 @@ if ( NOT NSTATIC )
 endif ( )
 
 # cholmod:
-target_link_libraries ( ParU PRIVATE SuiteSparse::CHOLMOD )
 target_include_directories ( ParU PUBLIC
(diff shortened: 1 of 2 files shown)


mmuetzel:
ParU doesn't need to link to the CHOLMOD library. It does include its header though.
Should be fixed now.

DrTimothyAldenDavis:
As long as ParU can find the cholmod.h header file, I see that the ParU library itself doesn't need to link against CHOLMOD itself.

So the find_package (CHOLMOD 5.0.0) is still there, so it can find the right cholmod.h, but then the linking can skip CHOLMOD, right?

mmuetzel:
> So the find_package (CHOLMOD 5.0.0) is still there, so it can find the right cholmod.h, but then the linking can skip CHOLMOD, right?

Exactly. That is needed when building ParU before CHOLMOD is installed.

PR #489, opened 2023-11-01
CXSparse: Avoid overlinking

mmuetzel:
CXSparse doesn't call a function from the suitesparseconfig library. It only includes its header for some definitions.

Don't link to the library to avoid overlinking.

### DIFF
--- a/CXSparse/CMakeLists.txt
+++ b/CXSparse/CMakeLists.txt
@@ -145,15 +145,11 @@ endif ( )
 # add the library dependencies
 #-------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


DrTimothyAldenDavis:
Good point but it reminds me that I was planning on revising CXSparse so that it uses the malloc etc functions in SuiteSparse_config.

CSparse will remain stand alone for simplicity since it's meant as an algorithmic teaching playground and CXSparse is what should be used in production. 

So I'd like to leave the link to the SuiteSparse_config library in place

mmuetzel:
Understood. 👍

Issue #474, opened 2023-10-29
ParU: demo intermittently stuck indefinitely


mmuetzel:
**Describe the bug**
When running `make demos` for ParU, execution occasionally is stuck indefinitely in `paru_demo`.

**To Reproduce**
Build ParU and run `make demos`.
It only happens occasionally. So, it might be a threading issue. I'll try to attach a debugger to the process when that happens the next time. Maybe that can give a clue where it is stuck.

**Expected behavior**
The demo executable terminates in a finite time.

**Desktop (please complete the following information):**
 - OS: Windows
 - compiler: gcc.exe (Rev2, Built by MSYS2 project) 13.2.0
 - BLAS and LAPACK library: OpenBLAS with OpenMP threading

DrTimothyAldenDavis:
We haven't tested ParU on Windows until now, if I recall.  This is using OpenMP 4.5 for ParU itself, correct?

@Aznaveh : We're making progress on getting the ParU cmake build system updated to the latest SuiteSparse, thanks to @mmuetzel's help.

One solution would be to disable OpenMP for ParU itself on Windows entirely, since it might take some time to track this down.  ParU would then use parallelism inside the BLAS and LAPACK alone, on Windows.  That's not ideal since ParU is meant to be able to factorize many frontal matrices in parallel, but this would at least be a stable temporary solution in order to get to a first stable release.

mmuetzel:
Seems to have happened in CI here:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/6711348727/job/18238519007?pr=486#step:9:6007

The job timed out after 6 hours.

mmuetzel:
Again here:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/6723919834/job/18275058763?pr=491#step:9:4251
and here:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/6736593096/job/18312283038#step:9:4253

mmuetzel:
Happened again: https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/6736593096/job/18312283038#step:9:4262

So far, this happened once for MINGW32 and three times for MINGW64 (afaict). It didn't happen on CLANG* or MSVC runners.
I wonder if that is already a pattern and what could be the reason for this.
[...]

PR #491, opened 2023-11-01
GraphBLAS: Generate compressed sources for JIT compiler automatically

mmuetzel:
This adds rules to automatically generate the compressed sources for the JIT compiler when building GraphBLAS.
No manual steps should be required for this any more.


mmuetzel:
The MINGW64 runner has timed out during the tests for ParU (see #474).
Other than that this seems to have worked. 🎉

PR #493, opened 2023-11-02
LAGraph: Add CMake config and pkg-config files

mmuetzel:
This adds generation of CMake Config files and pkg-config files for the LAGraph project.
It also restores the FindGraphBLAS module in LAGraph to be able to link with older versions of GraphBLAS (or potentially versions from different vendors). But I haven't actually tested that last part. I only built it with the version of GraphBLAS that is currently in SuiteSparse.

Like for ParU, I only added the part to the example that imports the targets and adds the links. But there is not actually a call to a function from LAGraph in the example. I would appreciate if someone else could add that.


mmuetzel:
Running `demo_paru` timed out again (on MINGW64), see #474. Other than that this seems to have built correctly. 👍

mmuetzel:
I'd like to include LAGraph in the root CMakeLists.txt for #415. That would require this PR to be merged first.
@DrTimothyAldenDavis: Could you please merge this PR when you come around to it?

PR #494, opened 2023-11-03
CI: Reduce timeout for check step

mmuetzel:
Running that step takes usually hardly longer than 5 minutes on native hardware. Set the timeout to 20 minutes to fail earlier if a runner gets stuck.
On emulated hardware, this step takes usually around 10 minutes. Set the timeout for this step to 30 minutes for the emulated runners.


DrTimothyAldenDavis:
Yes, this is a good idea, given that a successful check doesn't take much time.

PR #500, opened 2023-11-05
ParU: Use sequential factorization on Windows with GCC unconditionally

mmuetzel:
GCC with emuTLS on Windows might have trouble with shared linking. See #474.

I don't know if emuTLS is the actual reason why the demo gets stuck intermittently on the MINGW* runners. But this change is probably worth a try.

### DIFF
--- a/ParU/CMakeLists.txt
+++ b/ParU/CMakeLists.txt
@@ -162,6 +162,19 @@ if ( NOT NSTATIC )
 endif ( )
 
 #-------------------------------------------------------------------------------
+# additional definitions
+#-------------------------------------------------------------------------------
+
+if ( WIN32 AND CMAKE_C_COMPILER_ID STREQUAL "GNU" )
+    # Some parallelization levels don't work correctly with GCC on Windows
+    # (because of emuTLS?). Omit them.
+    target_compile_definitions (ParU PRIVATE PARU_GCC_WINDOWS )
+    if ( NOT NSTATIC )
+        target_compile_definitions (ParU_static PRIVATE PARU_GCC_WINDOWS )
+    endif ( )
+endif ( )
+
+#-------------------------------------------------------------------------------
 # add the library dependencies
 #-------------------------------------------------------------------------------
 
--- a/ParU/Source/paru_factorize.cpp
+++ b/ParU/Source/paru_factorize.cpp
@@ -341,6 +341,9 @@ ParU_Ret ParU_Factorize(cholmod_sparse *A, ParU_Symbolic *Sym,
     // execute the task tree
     //--------------------------------------------------------------------------
 
+#if ! defined ( PARU_GCC_WINDOWS )
+    // The parallel factorization gets stuck intermittently on Windows with GCC.
+    // Use the sequential factorization unconditionally in that case.
     if ((int64_t)task_Q.size() * 2 > Control->paru_max_threads)
     {
         printf ("Parallel:\n") ;    // FIXME
@@ -433,6 +436,7 @@ ParU_Ret ParU_Factorize(cholmod_sparse *A, ParU_Symbolic *Sym,
         }
     }
     else
+#endif
     {
         PRLEVEL(1, ("Sequential\n"));
         Work->naft = 1;


mmuetzel:
Linking ParU statically didn't help:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/6760811622/job/18374996302#step:9:6018

Trying with a static GraphBLAS library now.

mmuetzel:
Oops. The GraphBLAS_static CMake target seems to be broken.
Let me have a look at it...

See #501

DrTimothyAldenDavis:
I've merged in all of your other PRs except for this one.  Hopefully that will simplify things.  I also took out the "#define NPR" in ParU/Source/paru_internal.h, which turns off the ParU debug printf's.  I have a few printf's in the paru demos so it will show where ParU stalls (if it still does).

I also reset the default options for GraphBLAS back to their defaults.  I had set those to COMPACT on and NJIT on, to speed up the CI when testing ParU.  If the CI is slow, those 2 settings in GraphBLAS/cmake_modules/GraphBLAS_JIT_paths.cmake can be set back to ON for faster compile/tests.

mmuetzel:
Thanks for pushing those changes.

When it comes to this PR: I totally confused the dependencies of ParU and LAGraph. 🤦
But at least that lead to uncovering another bug.

I rebased the change on a current head, and also linked the other libraries for the demos statically. I hope that in particular UMFPACK might make a difference because it also uses OpenMP.
That still leaves a dynamic link to the OpenBLAS libraries on the affected runners. But let's see what will happen with the changes so far...

mmuetzel:
That didn't seem to have made a difference either.

I'll try building with Clang also in the MINGW* environments next.

mmuetzel:
In the last two CI runs none of the MINGW* runners got stuck during paru_demo. I don't know if that is already a pattern.
They got significantly slower though (during the tests of LAGraph for whatever reason).

The main change between the last two runs and the ones before was that Clang was used on those runners.
So, it might be an issue with GCC on MinGW indeed.

I'll skip the part about static linking and test again.

mmuetzel:
During the tests, I noticed that the tests of LAGraph took significantly longer while I disabled the factory kernels of GraphBLAS. Is the JIT compiler working correctly on Windows?

mmuetzel:
Back on topic: It looks like GCC on Windows has issues with some construct used in paru_demo. (Possibly emuTLS related?)
A possible "solution" could be to emit a warning during the configuration if we detect that GCC is used on Windows. Additionally, we could deactivate OpenMP in ParU for that configuration by default.
That warning could say that ParU doesn't support GCC on Windows with OpenMP. Also we could warn that ParU will be significantly slower without OpenMP.
If a user activates OpenMP nevertheless for that configuration, we could warn about the risk of possible deadlocks.

Would something like that be ok?

PS: This assumes that deactivating OpenMP will avoid the issue which isn't confirmed yet.

DrTimothyAldenDavis:
I think the simplest solution is to detect if gcc is used on windows, and then either disable OpenMP for ParU entirely, or at least set a -D flag that I can test inside ParU.

I know exactly where the failure is occurring.  It's inside paru_factorize, here in the main parallel factorization loop.  The code prints "stage 0" then hangs, and never gets to the "finalize permutation" printf, nor the "remaining" printf on line 415:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/11af347feb8b76b010aa5adf74f6ea5ea42d92cc/ParU/Source/paru_factorize.cpp#L377-L383

But this parallel phase has a "sequential" fallback:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/11af347feb8b76b010aa5adf74f6ea5ea42d92cc/ParU/Source/paru_factorize.cpp#L435-L452

There is a single test to decide which method to use:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/11af347feb8b76b010aa5adf74f6ea5ea42d92cc/ParU/Source/paru_factorize.cpp#L340-L346

All I would need to do is force the "sequential" factorization if the -D flag is set, where one front is factorized at a time.  Each front will still be able to use OpenMP, inside the BLAS calls and assembly and such.

ParU uses nested parallelism when doing multiple fronts in parallel, or just single-level parallelism if not enough threads are available.  The latter case is the "sequential" case.

This will reduce parallelism but will still be able to exploit some parallel computation.  Just one front at a time.

Can the cmake detect that case that gcc is used on Windows, and set a -D flag for ParU accordingly?
Something like "-DPARU_GCC_WINDOWS" would be a good idea, so it's clear in the code that the #define is ParU-specific.

DrTimothyAldenDavis:
Looks perfect!

mmuetzel:
> Looks perfect!

Thanks. Your description was pretty good. So, easy to follow. 👍

I hope that resolves the issue.

I pushed first and then rebased to have it run in CI at least twice... (Edit: It looks like it didn't trigger the CI again after the rebase. I don't know why...)

DrTimothyAldenDavis:
I can go ahead and merge it in. Looks ready to me. 

I might be able to reduce some of the analysis time by not creating the parallel stages. But that's not a big issue and can wait until later. There are other things that I need to polish in ParU first.

mmuetzel:
I let the CI run twice on my fork. And the MINGW* runner didn't time out in either run. (I still don't know why the CI didn't trigger here.)

Marked as ready for review.

PR #504, opened 2023-11-06
LAGraph: workaround for strange handling of integer constant in MSVC

DrTimothyAldenDavis:
MSVC doesn't handle typecasting of the literal INT32_MIN (-2147483648) to double, in a list of expected test outputs.  The value becomes +2.147...e9.  The value is fine if changed to -2147483648.0 instead, so it's treated as a double precision value.


mmuetzel:
Just for clarification if someone comes across this in the future: The issue is with extreme integer literals (not constants). In particular, a literal with the "value" of  `INT_MIN` isn't converted correctly to a `double`.

DrTimothyAldenDavis:
Yes, that's the issue.  It's easy to work around, but a surprising problem to encounter.  I would expect that any C11-compliant compiler could safely typecast the literal value -2147483648 to double ... but I guess not.

PR #511, opened 2023-11-08
UMFPACK: Avoid non-POSIX flags for `diff` in `make demos` target

mmuetzel:
The flag `--strip-trailing-cr` is a non-standard flag to `diff` and isn't supported by that application on Alpine Linux. Try to use commands that work across POSIX-compatible platforms.

I also opted to use `diff -u` because I'm more used to reading git diffs. I hope it's ok to change the format of the output.


DrTimothyAldenDavis:
At some point I need to turn all these short demos into cmake tests.

PR #513, opened 2023-11-10
GraphBLAS: Support jitifyer without CMake for MinGW

mmuetzel:
These modifications make the JIT compiler work without CMake for me using MinGW compilers (from MSYS2).
I don't know enough about MSVC. So, I left it at using CMake.

With those changes, I see the following timings for the ctests of LAGraph:
* first run: Total Test time (real) = 170.06 sec
* second run: Total Test time (real) =  17.82 sec

Compared to without the patch:
* first run: Total Test time (real) = 647.04 sec
* second run: Total Test time (real) =  19.04 sec

So, it significantly speeds up the first run. (The difference in the respective second runs is probably "random". Maybe caused by different load or alike.)

I believe this should still work on Linux or macOS without any noticeable difference. But I didn't actually try.

Some of these changes are pretty MSYS2-specific (e.g., fixing relocation issues). So, I'd understand if you prefer not to take it (or only take parts?). In that case, I'd propose a similar patch to MSYS2 that could be applied downstream when building GraphBLAS for distribution.
It would be easier though if MSYS2 wouldn't need to maintain a separate patch.

### DIFF
--- a/GraphBLAS/Config/GB_config.h
+++ b/GraphBLAS/Config/GB_config.h
@@ -15,7 +15,7 @@
 
 // GB_C_COMPILER: the C compiler used to compile GraphBLAS:
 #ifndef GB_C_COMPILER
-#define GB_C_COMPILER   "/usr/bin/cc"
+#define GB_C_COMPILER   "/mingw64/bin/cc.exe"
 #endif
 
 // GB_C_FLAGS: the C compiler flags used to compile GraphBLAS.  Used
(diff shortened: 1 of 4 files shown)


DrTimothyAldenDavis:
Looks great!  I'll merge it in and then give it a try on my various Linux systems and Macs (I have both an Intel Macbook and an M1 Macbook).

The LAGraph ctests can still pass even if there's a compiler error.  If the kernel doesn't compile, I assume it's some sort of error in the user-defined type or operator.  Or it might be a broken flag or something.  If that happens, I can still go forward and do the work in the generic kernel instead, and I turn off the JIT so the compiler errors don't keep happening, as a fallback.

So I'll need to check it out carefully to make sure it doesn't break on Linux/Mac.  It looks fine, but if something comes up I can always #ifdef it back to the original for Linux/Mac, and this your updates for MinGW / MSYS2.

PR #514, opened 2023-11-10
Replace many C-style casts

gruenich:
Replace C-style casts by C++'s static_casts.
    
static_cast has type checking which helps to find mistakes.
Only for the cases where static_cast is ok, other cases might profit from further investigation.
Found by cppcheck.


DrTimothyAldenDavis:
looks great -- thanks!

PR #515, opened 2023-11-11
[ParU] Use more C++

gruenich:
Replace C code by C++ code.
- Use range-based for loops.
- Don't use own definition of NULL and ASSERT
- Use std::min and std::max

If you are unhappy with some of the commits, I can remove them and create separate pull requests.

### DIFF
--- a/ParU/Source/paru_internal.hpp
+++ b/ParU/Source/paru_internal.hpp
@@ -311,7 +311,7 @@ inline int32_t control_nthreads (ParU_Control *Control)
     }
     else
     {
-        return std::min(max_threads, static_cast<int32_t>(Control->paru_max_threads));
+        return std::min(max_threads, Control->paru_max_threads);
     }
 }
 
--- a/ParU/Source/paru_internal.hpp
+++ b/ParU/Source/paru_internal.hpp
@@ -138,11 +138,6 @@ static int print_level = 0;   // FIXME
 // define FALSE 0
 #define IMPLIES(p, q) (!(p) || (q))
 
-// NULL should already be defined, but ensure it is here.
-#ifndef NULL
-    #define NULL ((void *)0)
-#endif
-
 #define Size_max ((size_t)(-1))  // the largest value of size_t
 
 // internal data structures
--- a/ParU/Source/paru_internal.hpp
+++ b/ParU/Source/paru_internal.hpp
@@ -132,8 +132,6 @@ static int print_level = 0;   // FIXME
 // basic macros
 // -----------------------------------------------------------------------------
 
-#define MIN(a, b) (((a) < (b)) ? (a) : (b))
-#define MAX(a, b) (((a) > (b)) ? (a) : (b))
 #define EMPTY (-1)
 // already defined in amd
 // define TRUE 1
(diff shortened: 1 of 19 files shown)


ParU/Source/paru_internal.hpp
```diff
@@ -318,7 +316,7 @@ inline int32_t control_nthreads (ParU_Control *Control)
     }
     else
     {
-        return (MIN (max_threads, Control->paru_max_threads)) ;
+        return std::min(max_threads, static_cast<int32_t>(Control->paru_max_threads));
```
mmuetzel: That probably doesn't make a big difference here. But to avoid wrapping by the cast to a smaller type, it is generally safer to cast to the larger type in comparisons:
``` suggestion
        return std::min(static_cast<int64_t>(max_threads), Control->paru_max_threads);
```

gruenich: Your suggestion does not work, as `std::min` will return some `int64_t` value, but the function signature declares the return value as `int32_t`. Unless the function signature is changed, we cannot avoid the down-cast.  
What way do you prefer?

I like this change, as it demonstrates the advantages of C++'s type system. Many type-related decisions have to be made explicit.

mmuetzel: You are absolutely right. The wrapping occurs in any case. So, this change doesn't make any difference.
The only (very weak) advantage might be that if the return type is changed to `int64_t` at some point in the future, this line wouldn't need to be changed.
On the other hand, 32-bit comparisons might be slightly faster than 64-bit comparisons on 32-bit platforms. So, maybe best leave it as it is. 👍

mmuetzel:
These changes look good to me.

DrTimothyAldenDavis:
Thanks for the update.  A few comments:

I can't use a bare `assert(...)`.  I always must wrap it inside a macro, for several reasons.  First, when used in MATLAB, the ASSERT macro must become a call to the MATLAB mxAssert function, not a C / C++ assert.  Second, I put computationaly expensive things inside my ASSERT(...) calls, and these must be disabled at all times. I enable them only by editting the code itself:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/19f11097c5b92ee6308d3eb48baa7fe272e865ac/ParU/Source/paru_internal.hpp#L21-L26

That is, I intentionally make it very hard to turn on debugging, since if it gets turned on by mistake by an end user, then my codes can be 1000x slower, or worse.  Then when I release the code, I force debug off by editing the code back.

And then sometimes I will revise the ASSERT macro so that it does something else, with a more informative error message, while developing the code.

DrTimothyAldenDavis:
Yes, in general, `std::min` and `std::max` would probably be better.  I would guess that the compiler might handle it better, in terms of performance.

The Control->paru_max_threads should be a plain int (or perhaps int32_t) anyway.  That would simplify this particular case for std::min and std::max.  This isn't a performance issue in this case, though.

For the for loops -- do they affect how OpenMP handles the loop?  I haven't reviewed all the changes yet, though.  I haven't used that style of for loop for a `#pragma omp parallel for` before.

DrTimothyAldenDavis:
looks great -- thanks!

PR #529, opened 2023-11-25
UMFPACK: Remove workaround for demos on POSIX platforms

mmuetzel:
It's no longer needed since we are linking to CMake targets instead of library names.


DrTimothyAldenDavis:
I'll merge this in once the current tests finish and are all green.

PR #531, opened 2023-11-26
CHOLMOD: Fix running ctests with only static libraries or on Windows

mmuetzel:
This works for me locally. Not yet tested in CI.


DrTimothyAldenDavis:
Thanks!  As you can see, I've made some major changes to CHOLMOD, with a full implementation of single precision sparse matrices.  I also took the opportunity to add ctests for CHOLMOD, but got stuck on the Windows tests.

DrTimothyAldenDavis:
Yay!

Issue #541, opened 2023-11-28
GraphBLAS doesn't export rmm_wrap when CUDA is enabled


DrTimothyAldenDavis:
@mmuetzel : I tried to enable cuda in SuiteSparse/GraphBLAS by setting ENABLE_CUDA to ON in SuiteSparse/GraphBLAS/CMakeLists.txt, but the cmake reports an error about not finding rmm_wrap.  I realize that you can't test the CUDA case on your end, but here's what I'm seeing.  Is this something that needs to be added to GraphBLAS/rmm_wrap/CMakeLists.txt?

```
hypersparse $ pwd
/home/faculty/d/davis/dev2/SuiteSparse/GraphBLAS
hypersparse $ make
( cd build && cmake  .. && cmake --build . --config Release -j8 )
-- Building SuiteSparse:GraphBLAS version: v8.3.0, date: Dec 30, 2023
-- GraphBLAS C API: v2.0, date: Nov 15, 2021
-- Source:        /home/faculty/d/davis/dev2/SuiteSparse/GraphBLAS 
-- Build:         /home/faculty/d/davis/dev2/SuiteSparse/GraphBLAS/build 
-- Install lib:      lib
-- Install include:  include
-- Install bin:      bin
-- Install pkg-file: lib
-- Install rpath:    
-- Build   rpath:    /home/faculty/d/davis/dev2/SuiteSparse/GraphBLAS/build
-- Build type:    Release 
-- Fortran: /usr/bin/f95
-- Looking for a CUDA compiler
-- Looking for a CUDA compiler - /usr/local/cuda/bin/nvcc
-- Looking for CUDA
-- Find CUDA tool kit:
-- Found CUDAToolkit: /usr/local/cuda/include (found version "11.7.99") 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- CUDA toolkit found:   TRUE
-- CUDA toolkit version: 11.7.99
-- CUDA toolkit include: /usr/local/cuda/include
-- CUDA toolkit lib dir: /usr/local/cuda/lib64
-- The CUDA compiler identification is NVIDIA 11.7.99
[...]

PR #543, opened 2023-11-28
GraphBLAS/rmm_wrap: Slightly overhaul build rules

mmuetzel:
Rename target to `RMM_wrap` and export it with the GraphBLASTargets.
Find package CUDAToolkit and link to CMake target instead of CUDA runtime library name.
Fix error when configured with `ENABLE_SHARED_LIBS=OFF`.

I can't test this change locally (no working CUDA yet). But I hope it will help for #541.

### DIFF
--- a/GraphBLAS/CUDA/CMakeLists.txt
+++ b/GraphBLAS/CUDA/CMakeLists.txt
@@ -243,26 +243,26 @@ set_target_properties(graphblascuda_test PROPERTIES CUDA_ARCHITECTURES "52;75;80
 
 include(GoogleTest)
 
-add_dependencies(graphblascuda_test GraphBLAS)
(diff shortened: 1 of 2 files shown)


DrTimothyAldenDavis:
Thanks!  I'll merge it in and test it out, both here and in the draft GraphBLAS 9.0.1 on the GraphBLAS repo.

DrTimothyAldenDavis:
Works!  

I had to make some other fixes, by back-porting those fixes from the draft GraphBLAS 9.0.1.  But this gets me back to where I was, in both v8.3.0 and v9.0.1.

I'll post those to v8.3.0 here, even though the CUDA kernels are fully functional yet.  But at least now the CUDA and RMM_WRAP libraries both build OK in v8.3.0.

thanks!

PR #548, opened 2023-11-29
LAGraph: Also look for GraphBLAS import target in common build tree

mmuetzel:
That is needed when using the root CMakeLists.txt.

### DIFF
--- a/LAGraph/cmake_modules/FindGraphBLAS.cmake
+++ b/LAGraph/cmake_modules/FindGraphBLAS.cmake
@@ -65,7 +65,7 @@ in your CMakeLists.txt file.  See also SuiteSparse/Example/CMakeLists.txt:
 ## New versions of SuiteSparse GraphBLAS (8.0.3 and newer) ##
 
 find_package ( GraphBLAS ${GraphBLAS_FIND_VERSION} CONFIG
-    PATHS ${PROJECT_SOURCE_DIR}/../GraphBLAS/build NO_DEFAULT_PATH )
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Doesn't this have to happen for all packages?  There could be a system amd.h or cholmod.h for example, and my other packages that depend on those mist link against the libraries and include files in the new build tree. Those must take precedence over any other system amd.h, cholmod.h etc

DrTimothyAldenDavis:
Or is that already in place but LAGraph is different because might need to look for other GraphBLAS implementations?

mmuetzel:
> Doesn't this have to happen for all packages? There could be a system amd.h or cholmod.h for example, and my other packages that depend on those mist link against the libraries and include files in the new build tree. Those must take precedence over any other system amd.h, cholmod.h etc

If I understand correctly, that shouldn't be a problem for the other libraries. They already have rules that prefer the libraries in the common source or build tree.

It's only the special logic in LAGraph that has more rules to find different versions of GraphBLAS that needs this change afaict.

DrTimothyAldenDavis:
Got it.  Yes, that makes sense.  Glad we caught this issue before SuiteSparse 7.4.0.beta1 was formally released.  Thanks for the fix!

mmuetzel:
> Glad we caught this issue before SuiteSparse 7.4.0.beta1 was formally released.

Sorry this slipped through.
It probably even picked up the GraphBLAS installed on my system during my testing without me noticing (because it is new enough). To look at the positive sight: That likely means that that CMake module *is* capable of importing different versions of GraphBLAS. 😉
But I'm also glad that this showed up during @Fabian188's tests. Took me a while to understand what was going on though...

PR #557, opened 2023-11-30
Remove dependencies for CUDA::nvtx3 and fix gpu.sh for cholmod

syby119:
I did the following change to the project
+ Remove CUDA::nvtx3
+ fix gpu.sh in CHOLMOD/Demo

It's quite embarrassing to admit that with cmake 3.22, the build system cannot find CUDA::nvtx3, as I test build with CUDA on WSL. Howerver, I find that the current project is unnecessary to link to the library. So I think remove the dependency for CUDA::nvtx3 is the best solution. Otherwise, we need to check whether CMake version is 3.25+. If it's above 3.25, then we can link to CUDA::nvtx3, elsewise we can link to CUDA::nvToolsExt.

Besides ,the script in CHOLMOD/Demo should be changed to cholmod_dl_demo instead of cholmod_l_demo. As the demo has specified the precision.
```
./gpu.sh: line 2: ../build/cholmod_l_demo: No such file or directory
./gpu.sh: line 3: ../build/cholmod_l_demo: No such file or directory
```


mmuetzel:
Looks good to me. 👍
I did almost the same changes in my local testing. (Trying to get CI for MSVC with CUDA. It's almost there. 🎉 Just waiting for #556.)

Could you please change the target branch to `dev2`?

PR #562, opened 2023-12-02
UMFPACK: Try fixing issues with old installed headers.

mmuetzel:
Avoid `include_directories` for system headers.
Link to libraries with `PRIVATE` scope also for static library if applicable.
Also remove the remaining parts for the workaround that was reverted in #529.
Additionally, cleanup dependencies and revert the debugging output.

@Fabian188 or @DrTimothyAldenDavis: Could you please check if this solves the compilation error of UMFPACK for you?


DrTimothyAldenDavis:
Works!

Issue #558, opened 2023-11-30
Install headers in a `suitesparse` subdirectory?


mmuetzel:
Motivated by build failures on macOS if old versions of SuiteSparse from Homebrew are installed when building SuiteSparse.
Copying a comment from [another thread](https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/175#issuecomment-1834265137) to avoid loosing this:

@DrTimothyAldenDavis: We can't do anything about older versions. But could the headers of the SuiteSparse libraries be installed in a sub-directory (e.g., /usr/local/include/suitesparse)?
Currently, there is the risk of headers from older libraries being picked up when the include path (e.g., /usr/local/include) gets added for other dependencies installed on the build system (e.g., OpenMP). That risk would be lower if the headers were installed in their own directory. CMake targets and pkg-config files would help dependent projects to find the correct folder with the headers.

Probably not something that should be changed for the current beta (or ever). But maybe for a later release?

DrTimothyAldenDavis:
That would just delay the problem.

The cmake build system knows where all the latest files are; if it's building AMD, let's say, then it knows that the correct SuiteSparse_config.h is in ../SuiteSparse_config/SuiteSparse_config.h.

Directories scanned with -I take precedence over system directories, right?  So we just need to make sure -ISuiteSparse_config/Include (for example) appears in all packages that use SuiteSparse_config.  Doesn't that always happen?

mmuetzel:
Yes, you are right. If this works it would fail later on for other libraries. I just wanted to see if this changes fixes the error for CHOLMOD on @Fabian188's system before making similar changes elsewhere.
If it doesn't allow building CHOLMOD for him, we might not need to make these kinds of changes at all ...

mmuetzel:
> Directories scanned with -I take precedence over system directories, right? So we just need to make sure -ISuiteSparse_config/Include (for example) appears in all packages that use SuiteSparse_config. Doesn't that always happen?

Those flags should be added for all libraries that use SuiteSparse_config currently. But they might be appearing after the -I flag for the other folder if the dependency for OpenMP is added earlier...

If the SuiteSparse headers hadn't been present in a folder that is shared with other (system) libraries, the order of the -I flags wouldn't matter.

We can try and control the order of the -I flags when building SuiteSparse. But if multiple versions of the headers are installed in different directories (like for @Fabian188) other projects that depend on SuiteSparse libraries could still get the wrong order of -I flags.
[...]

PR #568, opened 2023-12-03
Install SuiteSparse headers in subdirectory by default

mmuetzel:
This change would install (most of) the headers of SuiteSparse libraries to a subfolder `suitesparse` of the include folder.
That would prevent accidental inclusion of older SuiteSparse headers when the system include directory is added to the preprocessor search path. See #558.

### DIFF
--- a/SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake
+++ b/SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake
@@ -166,6 +166,9 @@ if ( NOT SUITESPARSE_ROOT_CMAKELISTS )
     endif ( )
 endif ( )
 
+set ( SUITESPARSE_INCLUDEDIR_POSTFIX "suitesparse" CACHE STRING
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
I will give it a try.  I can see that this could help avoid conflicts with /usr/local/include (my omp.h is there).  So this is probably a good thing to add.  If a package manager wants to move them elsewhere, they can do that themselves.

But if omp.h is in /opt/homebrew/include, alongside a stale amd.h from "brew install suite-sparse", then that will still cause a problem.

mmuetzel:
> I will give it a try. I can see that this could help avoid conflicts with /usr/local/include (my omp.h is there). So this is probably a good thing to add. If a package manager wants to move them elsewhere, they can do that themselves.
> 
> But if omp.h is in /opt/homebrew/include, alongside a stale amd.h from "brew install suite-sparse", then that will still cause a problem.

Yes, this won't fix it when the installed SuiteSparse headers are already mixed with the headers from other libraries. But it might make life easier at some point in the future.

Issue #565, opened 2023-12-02
SuiteSparse build system is finding stale header files of older SuiteSparse versions


DrTimothyAldenDavis:
@mmuetzel and @Fabian188  : We've had some discussions here: https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/561 on this topic, but that's a closed PR, and it would easier to discuss it here.

I'm still hitting this issue with Markus' latest PR ( https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/564 ).  

The top level SuiteSparse CMakeLists.txt seems fine (at least it builds the libcholmod.so).  Compiling cholmod_amd (I added linebreaks to make it easier to read):

```
[  2%] Building C object CHOLMOD/CMakeFiles/CHOLMOD.dir/Cholesky/cholmod_amd.c.o
cd /Users/davis/dev2/SuiteSparse/build/CHOLMOD &&
/Library/Developer/CommandLineTools/usr/bin/cc -DBLAS_Apple -DCHOLMOD_EXPORTS
-I/Users/davis/dev2/SuiteSparse/build/CHOLMOD
-I/Users/davis/dev2/SuiteSparse/CHOLMOD
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/Check
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/Cholesky
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/Utility
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/MatrixOps
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/Modify
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/Partition
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/Supernodal
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/Include
-I/Users/davis/dev2/SuiteSparse/SuiteSparse_config
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/SuiteSparse_metis/GKlib
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/SuiteSparse_metis/libmetis
-I/Users/davis/dev2/SuiteSparse/CHOLMOD/SuiteSparse_metis/include
-I/Users/davis/dev2/SuiteSparse/AMD/Include
-I/Users/davis/dev2/SuiteSparse/COLAMD/Include
[...]

PR #569, opened 2023-12-03
CMake: Don't include folders of imported targets with `SYSTEM` property.

mmuetzel:
Use `SYSTEM` explicitly only for the directory with OpenMP headers needed by SuiteSparse_config.

Might help for the issue discussed in #565.

@DrTimothyAldenDavis: Does it still build with this change? (With your OpenMP, and with SuiteSparse installed from Homebrew.)

Could you please show a compiler command for an object of SPEX when using the top-level Makefile (i.e., `make` in the root directory)?

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -92,6 +92,10 @@ if ( BUILD_SHARED_LIBS )
         PUBLIC_HEADER "Include/amd.h"
         WINDOWS_EXPORT_ALL_SYMBOLS ON )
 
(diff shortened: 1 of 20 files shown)


mmuetzel:
Oops. Probably got that wrong for the root CMakeLists.txt....

DrTimothyAldenDavis:
Thanks!  Can't try it just now but will do so later today.

mmuetzel:
We must have crossed by only a couple of seconds.

I opened #570 with the part that is missing from this PR. (Sorry for not noticing earlier.)

Issue #571, opened 2023-12-03
Mongoose tests are failing (various reasons)


DrTimothyAldenDavis:
The Mongoose tests have been revised, and they now work when building Mongoose inside Mongoose/build, but they fail when using the root CMakeLists.txt.  I had to make these 3 changes to get them to work in Mongoose/build:

1. This one I had to do because python3.12 broke the tests (distutils removed): https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/0ead217e95c74b10d53e6f9ecb221b86bfe74adf
2. This change was because the cmake target name changed at some point from mongoose_lib_dbg to Mongoose_static_dgb. The new name is better but then the Mongoose tests since then were silently failing: https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/afe685bb8039e3bfdad5a942391503f5a083afc2
3. This change was needed because if the directory "mongoose_lib_dbg" was not found, the tests fail but cmake thinks they pass: https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/f8ec2c6415204767f3a6e42ec59c59d4049a00d9

These issues were discussed in the wrong place: https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/565 .

However, there still is an issue.  The root CMakeLists.txt builds all of SuiteSparse inside SuiteSparse/build, and as a result, the Mongoose/Tests/runTests cannot find directories in needs.  An extra "Mongoose" or "../Mongoose" needs to be added to the many paths referenced in the tests.  For example "../Source" is refered to but it would need to be "../../Mongoose/Source" if the test is run inside SuiteSparse/build instead of Mongoose/build.

Then there are reference to Mongoose_static_dbg.dir, which is also no longer in the "right place" if the tests are run inside SuiteSparse/build.

I suppose the solution is to detect where the test are being run (in SuiteSparse/build or SuiteSparse/Mongoose/build) and to revise the file locations accordingly.  Not a hard change but there are many places where the paths need to be revised.

In the meantime, the Mongoose tests will be failing when using the root CMakeLists.txt.

DrTimothyAldenDavis:
[...]

PR #573, opened 2023-12-04
Mongoose: Fix ctests with MSVC

mmuetzel:
Should be fixing the remaining issue described in #571.

This is not a perfect solution and I left a FIXME note describing the shortcomings. But it should be good enough to have the ctests be running in the CI.


DrTimothyAldenDavis:
Great -- thanks!

Issue #572, opened 2023-12-04
GraphBLAS JIT isn't working in MINGW when using cmake


DrTimothyAldenDavis:
The GraphBLAS JIT is encountering compiler errors when compiling the JIT kernels on MINGW when using cmake.   The diagnostics are pretty terse but this occurs when GraphBLAS tries to use cmake to build a kernel.  The name configured by cmake to create GraphBLAS/Config/GB_config.h must be inserting this line:

`#define GB_C_COMPILER " /clang64/bin/cc.exe"
`

The failed output is below.  The gauss_demo still succeeds, because of how I handle compiler errors in the JIT.  If a JIT kernel fails to compile or load, it's hard to tell inside GraphBLAS exactly what the problem is.  It might be a syntax error in the user-defined operator, for example.  Since I have a function pointer anyway, I can fallback to using a "generic" kernel that uses this function pointer, and then I return the correct result to the user application.  I turn off the JIT in that case to avoid repeated errors.  So everything still works, just slower with no JIT.

I should revise my tests to assert that after compiling a valid kernel (like this one), that the JIT is still enabled.  Then I would have caught this error as a test failure.

The GB_C_COMPILER string is set by GraphBLAS/cmake_modules/GraphBLAS_JIT_configure.cmake, and there's a special case there for MINGW.

It's possible that the GB_COMPILER string is OK but this line is causing the problem for MINGW:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/0b41a3e842271e1b180447f42a185f3469e7d21d/GraphBLAS/Source/GB_jitifyer.c#L2268-L2276

Here's the log with the error.  I will add some more diagnostics to try to track this down (I can print out all my calls to `system(...)` for example:

```
  CMake Error at CMakeLists.txt:2 (project):
    The CMAKE_C_COMPILER:
  
      /clang64/bin/cc.exe
  
    is not a full path to an existing compiler tool.
  
    Tell CMake where to find the compiler by setting either the environment
    variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
    the compiler, or to the compiler name if it is in the PATH.
  
  
  ninja: error: loading 'build.ninja': The system cannot find the file specified.
  
  
  CMake Error: Error processing file: C:/msys64/home/runneradmin/.SuiteSparse/GrB8.3.0/tmp/ce2be202506bf1b4/cmake_install.cmake
  gauss_demo: all tests pass
```

[...]

PR #574, opened 2023-12-04
GraphBLAS: Run `cmake` in POSIX shell for MinGW environments.

mmuetzel:
Running the GraphBLAS demo `demo_gauss` still fails for the MinGW runners with this change. But I don't claim to understand what it actually checks...

See also #572.


DrTimothyAldenDavis:
I'll dig into the gauss_demo for MINGW some more.

PR #575, opened 2023-12-05
SuiteSparse 7.4.0.beta1

DrTimothyAldenDavis:
Dec 30, 2023: version 7.4.0.beta1

    * major change to build system: by Markus Mützel.  Includes a
        top-level CMakeLists.txt that builds all packages, and support for
        pkg-config.  Additional changes are listed below. 
    * LAGraph 1.1.0: new package: graph algorithms based on GraphBLAS
    * ParU 0.1.0: new package: parallel unsymmetric multifrontal method,
        with Mohsen Aznaveh.  This is a stable package but is tagged as 0.1.0
        since the API is still subject to change.
    * CHOLMOD 5.1.0: full support for sparse single precision matrices
    * AMD 3.3.0: minor change for CHOLMOD 5.1.0 tests
    * CAMD 3.3.0: minor change for CHOLMOD 5.1.0 tests
    * SuiteSparse_config 7.4.0: added wrappers for single-precision BLAS/LAPACK
    * *_version: added methods to all package that didn't have them:
        AMD, CAMD, COLAMD, CCOLAMD, BTF, CSparse, CXSparse, KLU, BTF, RBio,
        SPEX, SPQR, and UMFPACK.


DrTimothyAldenDavis:
Here we go!  I think I can release SuiteSparse 7.4.0.beta1 soon.  @mmuetzel, let me know if you see anything that still might need fixing.

mmuetzel:
Just stumbled upon some inconsistencies with the usage of the CMake variable `OpenMP_C_FOUND` (and related variables).
I'll open a PR shortly.

Edit: See #576

DrTimothyAldenDavis:
I had to make some minor changes to LAGraph/cmake_modules/FindGraphBLAS.cmake:

(1) The old version with `IMPORTED_INCLUDE_DIRECTORIES` doesn't work; it needs to be `INTERFACE_INCLUDE_DIRECTORIES`.  I made this change to get it to work for GraphBLAS 7.4 and earlier.

(2) The new target, GraphBLAS::GraphBLAS, exists in GraphBLAS 8.2.0 and 8.2.1, which are stable released versions.  But their `INTERFACE_INCLUDE_DIRECTORIES` is the top-level GraphBLAS folder, not "GraphBLAS/Include".  So I just added an extra `include_directories` as a workaround.  I know that isn't ideal, but this is an unusual combination.  LAGraph 1.1.0 in SuiteSparse 7.4.0 will be coupled with GraphBLAS 8.3.0, which doesn't need this workaround.  If GraphBLAS 8.2.x is installed in a known place, and LAGraph 1.1 is used from the stand-along LAGraph distro, the `INTERFACE_INCLUDE_DIRECTORIES` is probably fine (perhaps /usr/local/include) so this workaround is not needed for LAGraph 1.1 anyway, but then it probably doesn't hurt anything either.  I will need to test that to be sure before I merge this PR, so this is still a work in progress.

mmuetzel:
Trying to understand issue (2): Are you linking to GraphBLAS version 8.2.x that is built in a common build tree and not yet installed? Or does the issue also happen if that older GraphBLAS version was already installed?

DrTimothyAldenDavis:
Yes. I have a repo for GraphBLAS that is stand alone and the released version there is 8.2.1. I will update it soon with 8.3.0. 

8.2.x has your earlier cmake scripts you worked on for SuiteSparse 7.3. But the package config script has the wrong include directory: just the top level GraphBLAS folder. 

The workaround is not needed if 8.2.x is fully installed. It's only needed if GraphBLAS and LAGraph are placed side by side but not part of Suite Sparse itself, and if GraphBLAS is built but "make install " has not been done. 

This only affects GraphBLAS 8.2.0 and 8.2.1, but only if built but not installed. For the older ones, 7.4.x and earlier, the old style is used. They don't export the :: target. So they act like vanilla GraphBLAS libraries. 

GraphBLAS 8.3.0 has no problem with this. It has the right include directory in its config cmake scripts. 

So it's a very narrow workaround.

mmuetzel:
Thank you for clarifying. So, it's likely the issue that was fixed with https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/450/commits/e1c7e4ea0bc651c8136cc3a71fd8228bccb72644 and https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/2731ed9efd1955ef0a5be4b3183d5437379c6ac6.

DrTimothyAldenDavis:
The older GraphBLAS 7.4.x to 7.1x work fine with LAGraph 1.1, either installed or not.  They trigger the "vanilla" case in the LAGraph FindGraphBLAS.cmake script, and it works fine.

Here's what happens when using GraphBLAS 8.2.x and LAGraph 1.1, if GraphBLAS is not installed, and if these libraries are not inside SuiteSparse.  From the LAGraph FindGraphBLAS.cmake script:

```
...
-- Looking for SuiteSparse GraphBLAS
-- GraphBLAS version: 8.2.1
-- GraphBLAS include: /home/faculty/d/davis/v821/GraphBLAS
-- GraphBLAS library: /home/faculty/d/davis/v821/GraphBLAS/build/libgraphblas.so.8.2.1
-- GraphBLAS static:  
-- Found SuiteSparse::GraphBLAS
-- additional include: /home/faculty/d/davis/v821/GraphBLAS/Include
...
```

So the FindGraphBLAS.cmake finds the GraphBLAS::GraphBLAS target, using the cmake config scripts in LAGraph/../GraphBLAS/build.  But notice the report where it says `"-- GraphBLAS include: /home/faculty/d/davis/v821/GraphBLAS`".  That's a misconfiguration in the 8.2.1.  The file in GraphBLAS/build/GraphBLASTargets.cmake for 8.2.1 says the following; it's missing a trailing `/Include` in the path:

```
set_target_properties(SuiteSparse::GraphBLAS PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "/home/faculty/d/davis/v821/GraphBLAS"
)
```

For 8.3.0, it has it correctly defined as follows (this is from a build where I placed just LAGraph and GraphBLAS alone, in a directory outside of SuiteSparse, side-by-side):

```
set_target_properties(SuiteSparse::GraphBLAS PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "/home/faculty/d/davis/v830/GraphBLAS/Include"
)
```

I can't go back in time to fix 8.2.x; those are already released.  This is just a minor workaround for backward compatibility.  Normally, someone using LAGraph 1.1 would use the corresponding GraphBLAS 8.3.0 anyway, in the current SuiteSparse 7.4.0, and it all works fine.

But in the LAGraph repo, the current version of LAGraph is tested against a range of GraphBLAS versions (currently versions 7.1.0, 7.3.0, and 7.4.1).  That will be augmented with subsequent versions, but probably just the latest one, 8.3.0.   If 8.2.x is tried with LAGraph 1.1, however, it would fail unless 8.2.x is not just build, but also installed (and then the workaround is not needed).

DrTimothyAldenDavis:
> Thank you for clarifying. So, it's likely the issue that was fixed with [e1c7e4e](https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/e1c7e4ea0bc651c8136cc3a71fd8228bccb72644) and [2731ed9](https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/2731ed9efd1955ef0a5be4b3183d5437379c6ac6).

Yes, that's exactly it.  The older version, before this fix, appears in GraphBLAS 8.2.0 and 8.2.1.

I think my workaround is safe.  I guard it with `if ( GRAPHBLAS_VERSION LESS "8.3.0")`, and the entire block of code there is only triggered by GraphBLAS 8.2.0 and higher.

mmuetzel:
> > Thank you for clarifying. So, it's likely the issue that was fixed with [e1c7e4e](https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/e1c7e4ea0bc651c8136cc3a71fd8228bccb72644) and [2731ed9](https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/2731ed9efd1955ef0a5be4b3183d5437379c6ac6).
> 
> Yes, that's exactly it. The older version, before this fix, appears in GraphBLAS 8.2.0 and 8.2.1.
> 
> I think my workaround is safe. I guard it with `if ( GRAPHBLAS_VERSION LESS "8.3.0")`, and the entire block of code there is only triggered by GraphBLAS 8.2.0 and higher.

Your work-around looks good to me. I opened #578 to limit when the work-around is applied a bit more and add the include directory only in the scope of the targets.

DrTimothyAldenDavis:
I've tested the change with GraphBLAS 7.1.0 (the earliest version LAGraph allows), 7.3.0, 7.4.1, 7.4.4, 8.0.0, 8.0.1, 8.0.2, 8.2.0, and the current version in SuiteSparse/dev2: 8.3.0.

Versions 8.2.x fail with this error message below.  All other versions work fine:

```
-- Looking for SuiteSparse GraphBLAS
-- GraphBLAS version: 8.2.0
-- GraphBLAS include: /home/faculty/d/davis/v820/GraphBLAS
-- GraphBLAS library: /home/faculty/d/davis/v820/GraphBLAS/build/libgraphblas.so.8.2.0
-- GraphBLAS static:  
-- Found SuiteSparse::GraphBLAS
CMake Error at cmake_modules/FindGraphBLAS.cmake:94 (target_include_directories):
  target_include_directories can not be used on an ALIAS target.
  CMakeLists.txt:154 (find_package)
```

mmuetzel:
Oops. Aliased targets again...
I opened #579 with a potential fix (but untested).

DrTimothyAldenDavis:
Your fix is almost working; I had to make a minor tweak.

In one case on my Mac, I was trying to use GraphBLAS 8.0.2 (with no "::" namespace) with LAGraph 1.1.  It failed because I had a copy of "brew install suite-sparse" on the Mac, and LAGraph found that instead, as version 8.2.1.

That version is configured correctly with the include file as /opt/homebrew/include, and tacking on the additional /opt/homebrew/include/Include gives a warning message saying the directory doesn't exist.

So I simply added a check: only add the new target_include_directory if it actually exists.  See https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/575/commits/d7daee52668d9d8279b2abc4c6b12d48852bf4a3

And now it works fine.

This is an edge case, on one hand, but important because it's possible someone wants to grab LAGraph from its own repo,and use their existing "brew suite-sparse" version of GraphBLAS (v8.2.1).  That will work fine now.

I'll do some more testing on more platforms with LAGraph 1.1 and older versions of GraphBLAS, but so far it's working great (with this fix).

SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake
```diff
@@ -74,9 +74,20 @@
...
+#   SUITESPARSE_PKGFILEDIR: where to install the CMake Config and pkg-config
+#                       files.  This defaults to the same directory as where
+#                       the compiled libraries are installed, in a subfolder
+#                       called cmake.  If not already set in the cache,
+#                       LOCAL_INSTALL=0 defines this as "lib", and the
+#                       CMAKE_INSTALL_PREFIX is added.  LOCAL_INSTALL=1 defines
+#                       this as SuiteSparse/lib.  This variable is cached so
+#                       that if it is not set, or unset first, it remains
+#                       unchanged (see "make local" and "make global" in the
+#                       SuiteSparse_config/Makefile for an example).
```
mmuetzel, line 88: *All* CMake configuration flags are cached.

Is there a particular reason why this flag should be handled differently to all other configuration flags?

SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake
```diff
@@ -74,9 +74,20 @@
...
 #                       system.
 #                       Default: false
-
-message ( STATUS "Source:        ${CMAKE_SOURCE_DIR} ")
-message ( STATUS "Build:         ${CMAKE_BINARY_DIR} ")
+#
+#   SUITESPARSE_PKGFILEDIR: where to install the CMake Config and pkg-config
+#                       files.  This defaults to the same directory as where
+#                       the compiled libraries are installed, in a subfolder
+#                       called cmake.  If not already set in the cache,
```
mmuetzel, line 82: That is a bit confusing. The pkg-config files aren't installed into a subfolder called `cmake`. (At least, I hope they aren't. If they are, that would be an error.)

See the documentation of that flag in `README.md`.

SuiteSparse_config/Makefile
```diff
@@ -41,11 +41,11 @@ library:
 
 # install only in SuiteSparse/lib and SuiteSparse/include
 local:
-	( cd build && cmake $(CMAKE_OPTIONS) -DLOCAL_INSTALL=1 .. && cmake --build . --config Release -j${JOBS} )
+	( cd build && cmake $(CMAKE_OPTIONS) -USUITESPARSE_PKGFILEDIR -DLOCAL_INSTALL=1 .. && cmake --build . --config Release -j${JOBS} )
 
 # install only in /usr/local (default)
 global:
-	( cd build && cmake $(CMAKE_OPTIONS) -DLOCAL_INSTALL=0 .. && cmake --build . --config Release -j${JOBS} )
+	( cd build && cmake $(CMAKE_OPTIONS) -USUITESPARSE_PKGFILEDIR -DLOCAL_INSTALL=0 .. && cmake --build . --config Release -j${JOBS} )
```
mmuetzel, line 48: It would be surprising to me as a user if the build rules would undefine a flag that I set in `CMAKE_OPTIONS`.

Would it be better to undefine that flag *before* adding the flags from `CMAKE_OPTIONS`?

SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake
DrTimothyAldenDavis: I thought it had to be different.  It was added here:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/a02c53b1285ab7c0a128b1dc7b26ce2288281d39/SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake#L234-L235

outside the if/else part for other variables:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/a02c53b1285ab7c0a128b1dc7b26ce2288281d39/SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake#L193-L214

So I thought it must be different somehow.  But then when I did "make local" after "make", the variable did not get reset, and it was out of sync with the rest of the variables.

SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake
DrTimothyAldenDavis: That was a typo; I just fixed it.

SuiteSparse_config/Makefile
DrTimothyAldenDavis: I had trouble with this variable getting "stuck" in one mode and not getting changed to the correct value with "make local".

PR #576, opened 2023-12-05
Fix typos in CMake variable `OpenMP_C_FOUND`.

mmuetzel:
CMake variables are case-sensitive.
This fixes a couple of typos and inconsistencies when it comes to the CMake variable `OpenMP_C_FOUND` (and other similar ones).

### DIFF
--- a/GraphBLAS/CMakeLists.txt
+++ b/GraphBLAS/CMakeLists.txt
@@ -42,8 +42,8 @@ if ( NOT BUILD_SHARED_LIBS )
 endif ( )
 
 # CUDA is under development for now, and not deployed in production:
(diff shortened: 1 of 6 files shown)


DrTimothyAldenDavis:
Looks great.  Shall I call it 7.4.0.beta1?

I'll wait a while before a stable 7.4.0 release.  I tagged a tentative date of Dec 30, hopefully no later than that.  Maybe I can do it earlier.

So there's time for additional changes like this, if you see something else.

mmuetzel:
That sounds good to me. 👍

It would probably be good to get a couple more eyes on this to catch potential issues before a release. Tagging a beta would probably help with that.

Issue #478, opened 2023-10-31
CHOLMOD 5.0.0's header clashes with <complex.h>


UncombedCoconut:
This is not a bug, per se, but rather a bad interaction which may have been overlooked.
Suppose for some reason you've `#include`d `<complex.h>` -- directly or otherwise -- and later `<cholmod.h>`.
Then, under SuiteSparse 7.3.0 / CHOLMOD 5.0.0, you're dead because this declaration clashes with the `I` macro for the imaginary unit:
```C
int cholmod_realloc_multiple    // returns true if successful, false otherwise
(
    size_t nnew,    // # of items in newly reallocate memory
    int nint,       // 0: do not allocate I or J, 1: just I, 2: both I and J
    int xdtype,     // xtype + dtype of the matrix:
                    // (CHOLMOD_DOUBLE, _SINGLE) +
                    // (CHOLMOD_PATTERN, _REAL, _COMPLEX, or _ZOMPLEX)
    // input/output:
    void **I,       // integer block of memory (int32_t or int64_t)
    void **J,       // integer block of memory (int32_t or int64_t)
    void **X,       // real or complex, double or single, block
    void **Z,       // zomplex only: double or single block
    size_t *n,      // current size of I, J, X, and/or Z blocks on input,
                    // changed to nnew on output, if successful
    cholmod_common *Common
) ;
```

In your opinion, is it worth tweaking this declaration to avoid this problem? (I'll live if the answer is no.)
Thanks!

(Reproducer: with GCC 13 or Clang 16, run `cc -c` on a file with statements `#include <complex.h>` and `#include <cholmod.h>`. There's a compile error on the included `void **I,` line.)

mmuetzel:
Thanks for your report.
CHOLMOD should probably avoid to repurpose symbols that could be defined by including a standard header. Just renaming `I` in the header would probably be fine.

DrTimothyAldenDavis:
Oops.  In my opinion, C11 adding a single letter macro "I" was a really bad mistake, but I can't fix that ...

I'll fix it shortly and release a CHOLMOD 5.0.1 and SuiteSparse 7.3.1.

UncombedCoconut:
Thank you! Yes, I agree with that assessment. :)

DrTimothyAldenDavis:
[...]

PR #581, opened 2023-12-09
revise *.pc, undef I

DrTimothyAldenDavis:
GraphBLAS.h in v8.3.0 and earlier has "#undef I".  So did CHOLMOD (see https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/478).  This PR removes "#undef I" from GraphBLAS.h.

The *.pc files are revised to add the release postfix.


mmuetzel:
GraphBLAS shouldn't "lie" about the complex data types that it uses in its interface. The types for `GxB_FC32_t` and `GxB_FC64_t` shouldn't be different depending on whether the header happens to be included in a C or C++ source file (or in an `extern "C"` block)...


If GraphBLAS CUDA really uses a "different" type for complex numbers in its interface, it shouldn't use the same name for it then the rest of GraphBLAS.

DrTimothyAldenDavis:
The GraphBLAS cuda kernels do not yet work at all for the complex data type.  This first block is a temporary workaround and even then will only be triggered from inside GraphBLAS itself with an internally #defined flag. No user application will see that first block.

DrTimothyAldenDavis:
Correction:  that block (in GxB_complex.h at least) could be included in a user defined operator passed as a string to GraphBLAS. The user operator can then be compiled both in C in the CPU JIT and as a device function inside a CUDA jit kernel compiled by nvcc (the latter is not yet working in production). 

So one string will serve both cases. When compiling, the idea is that the string is given GBCUDA_CPLUSPLUS #defined if it's inside a cuda kernel. 

We may instead decide to allow a user operator to be defined with two strings:  one in C to use one the CPU and one in C++ to use as a CUDA device function. 

But none of that is in place yet. This is a future work and we never yet #define 
GBCUDA_CPLUSPLUS for any user code.

DrTimothyAldenDavis:
On 2nd thought, that bit of code in GraphBLAS.h *is* confusing, and it's likely to change when we get the CUDA kernels working in production anyway.  So I'll go ahead and delete it.

DrTimothyAldenDavis:
The new FindGraphBLAS.cmake works great.  I tested it with GraphBLAS 7.1.0 and it found it.  Works with 8.2.x as well where cmake reports this, as expected:

`-- additional include: /home/faculty/d/davis/v821/GraphBLAS/Include
`

PR #582, opened 2023-12-09
CMake: Account for changes to library base names by CMake flags

mmuetzel:
Library base names can be prefixed, suffixed, or otherwise changed by setting a range of CMake configuration flags.
Extract the base name of the library from the CMake target and use that name when generating the package config files.

For the libraries themselves that should be pretty straight forward.

For the name of the GraphBLAS library used when building LAGraph, this is a bit more tricky because the GraphBLAS library is allowed to come from different sources (with or without import targets).
Try to extract the name of the library from the information that is gathered in `FindGraphBLAS.cmake` and add it to the created target `GraphBLAS::GraphBLAS` also when the used GraphBLAS doesn't provide an import target.

I think the second part should be working. But I haven't actually tested it. (No old or different GraphBLAS here...)

See the issue described in this comment (and following ones): https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/425#issuecomment-1847641837

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -230,10 +230,19 @@ if ( NOT MSVC )
     else ( )
         set ( includedir "\${prefix}/${SUITESPARSE_INCLUDEDIR}")
     endif ( )
+    if ( BUILD_SHARED_LIBS )
+        set ( SUITESPARSE_LIB_BASE_NAME $<TARGET_FILE_BASE_NAME:AMD> )
+    else ( )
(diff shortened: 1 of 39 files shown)


DrTimothyAldenDavis:
I'll give it a try. On my phone just now so can't test it yet. I can try with GraphBLAS 7.1 which acts like a "vanilla " library since it triggers the 2nd part of FindGraphBLAS.cmake.

DrTimothyAldenDavis:
Works great! I pushed it to dev.

PR #591, opened 2023-12-14
installation locations

DrTimothyAldenDavis:


ChangeLog
```diff
@@ -2,7 +2,17 @@ Dec 30, 2023: version 7.4.0
 
     * major change to build system: by Markus Mützel.  Includes a
         top-level CMakeLists.txt that builds all packages, and support for
-        pkg-config.  Additional changes are listed below. 
+        pkg-config. Default location of files is now listed below, where
+        PACKAGE is one of the packages in SuiteSparse:
+        * CMAKE_INSTALL_PREFIX/include/suitesparse: include files
+        * CMAKE_INSTALL_PREFIX/lib: compiled libraries
+        * CMAKE_INSTALL_PREFIX/lib/cmake/SuiteSparse: *Config.cmake scripts
```
mmuetzel, line 9: Afaict, the .cmake files installed to that folder aren't `Config` files (which usually import CMake targets).

To avoid confusion, maybe that should be changed to:
``` diff
-        * CMAKE_INSTALL_PREFIX/lib/cmake/SuiteSparse: *Config.cmake scripts
+        * CMAKE_INSTALL_PREFIX/lib/cmake/SuiteSparse: helper *.cmake scripts
```

ChangeLog
```diff
@@ -2,7 +2,17 @@ Dec 30, 2023: version 7.4.0
...
+            for all of SuiteSparse
+        * CMAKE_INSTALL_PREFIX/lib/cmake/PACKAGE: *Config.cmake scripts for a
+            specific package
+        * CMAKE_INSTALL_PREFIX/lib/pkgconfig/PACKAGE.pc: *.pc scripts for
+            a specific package pkgconfig
```
mmuetzel, line 14: Maybe, the wording could be changed to something like:
``` diff
-        * CMAKE_INSTALL_PREFIX/lib/pkgconfig/PACKAGE.pc: *.pc scripts for
-            a specific package pkgconfig
+        * CMAKE_INSTALL_PREFIX/lib/pkgconfig/PACKAGE.pc: *.pc pkg-config
+            files with information for a specific package
```
(Those files aren't actually "scripts".)

DrTimothyAldenDavis: Yes - those are better -- I'll revise it now.

PR #593, opened 2023-12-15
revise changelog and CI

DrTimothyAldenDavis:


mmuetzel:
Just out of interest: What was the motivation for having so different CI rules for macOS compared to the other platforms?

DrTimothyAldenDavis:
Several reasons:  I was having trouble with the macos CI because the check phase was timing out at 20 minutes, and then even 40 minutes.  So to diagnose it, I split it out into its own macos.yaml CI.  The check phase was hanging at random places in that loop, and I couldn't tell where, so I removed the loop and put in hard coded rules, per package.

At one point, ParU was hanging on the Mac because of tasking in its innermost loop, so that got disabled.  I was concerned that I still had an intermittent hang in ParU, so I removed it from the check entirely.  That didn't solve the hang, and then I put it back.

Since the mac had trouble with the old CI, and this one is now working, I'd like to leave it separated like this in case the issue arises again.  I'd also like to get the Mac working with OpenMP tasking in ParU sometime, and for that it's best to have ParU in its own check phase.

The other reason I split the mac into its own yaml file is for caching.  The build.yaml rules finish fast if nothing changes ... except for the macos part.  It seems to want to recompile everything most of the time.  So if it's in its own macos.yaml file, then build.yaml finishes as green while the macos.yaml can then keep spinning.

mmuetzel:
Thank you for the explanation.
I also noticed that the CI got stuck intermittently when linking one of the demos in LAGraph.
You're no longer building the demos for LAGraph. So, I guess that's why that issue disappeared.

I don't know if removing problematic parts from the CI is the best usage of it though. It might be better to find out *why* linking one (or some?) of the demos of LAGraph causes issues sometimes...

mmuetzel:
> The other reason I split the mac into its own yaml file is for caching. The build.yaml rules finish fast if nothing changes ... except for the macos part. It seems to want to recompile everything most of the time. So if it's in its own macos.yaml file, then build.yaml finishes as green while the macos.yaml can then keep spinning.

At the moment, both rules use the same key for `concurrency`. So, the two rules are blocking each other instead of building concurrently.

PR #598, opened 2023-12-17
[ParU] Small performance improvement in update rowDeg

gruenich:
std::set only inserts an element, if it does not already contain the element. Thus, we can get rid of the additional find.

This was found by cppcheck.

No pull request can be accepted unless you first sign the Contributor License Agreement [ CONTRIBUTOR-LICENSE.txt ].  Print it as a PDF and email me a signed PDF (digital signature OK).  Submit all PRs to the dev2 branch only.


mmuetzel:
Looks good to me. 👍

DrTimothyAldenDavis:
Looks great -- thanks.

Issue #599, opened 2023-12-18
Improved CMake variables / behavior


haampie:
Hi, I'm trying out the new CMake build system after struggling to keep up with all breaking changes in the Makefile-based build system.

Few comments.

1. Could you please drop double negatives, it's only confusing. Do not not use openmp?

   ```cmake
   option ( NOPENMP "ON: do not use OpenMP.  OFF (default): use OpenMP" OFF )
   ```

2. Please prefix user facing variables: for example `SUITESPARSE_WITH_OPENMP`, `SUITESPARSE_WITH_CUDA`, etc.

3. I don't understand why there's so much special handling of rpaths in suite sparse? What problem is being solved that is not already solved by CMake? (e.g. `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON`) Is there anything unique about suitesparse that requires this? If not, could you please remove special handling and leave it to packagers?

4. Please use `find_package(... REQUIRED)` instead of silently turning a feature off.

    ```cmake
    if ( NOPENMP )
        # OpenMP has been disabled
        message ( STATUS "OpenMP disabled" )
        set ( OpenMP_C_FOUND OFF )
    else ( )
        find_package ( OpenMP )  # <--- find_package(OpenMP REQUIRED) would be better
    endif ( )
    ```
    Same for CUDA: if enabled by the user, require it, instead of
    
    https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/a02c53b1285ab7c0a128b1dc7b26ce2288281d39/LAGraph/cmake_modules/SuiteSparsePolicy.cmake#L301-L304

5. Please be clear about the minimum CMake version. The top-level CMakeLists.txt requires 3.20 but says higher may be required. What's the minimum then? I don't want to go over each and every subproject. Just set the top-level min to `max(min(subproject) for subproject in subprojects)`

   Further, there are a lot of outdated and redundant cmake policy things, can they be removed? They're all NEW by default.
   
   https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/a02c53b1285ab7c0a128b1dc7b26ce2288281d39/SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake#L98-L103

[...]

PR #601, opened 2023-12-18
Revert "Set RPATH for installed binaries"

mmuetzel:
This reverts commit 222b7ee17b9b2ce6543b5165171c74f7b641efb4.

See #599 and https://gitlab.com/petsc/petsc/-/merge_requests/7104#note_1699384612

### DIFF
--- a/GraphBLAS/cmake_modules/SuiteSparsePolicy.cmake
+++ b/GraphBLAS/cmake_modules/SuiteSparsePolicy.cmake
@@ -155,7 +155,7 @@ endif ( )
 # find this one without "make install"
 list ( FIND CMAKE_BUILD_RPATH ${CMAKE_BINARY_DIR} _idx )
 if ( _idx LESS 0 )
-    # not yet included in CMAKE_INSTALL_RPATH
+    # not yet included in CMAKE_BUILD_RPATH
     list ( APPEND CMAKE_BUILD_RPATH ${CMAKE_BINARY_DIR} )
 endif ( )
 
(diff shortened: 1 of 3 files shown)


GraphBLAS/cmake_modules/SuiteSparsePolicy.cmake
```diff
@@ -155,7 +155,7 @@ endif ( )
 # find this one without "make install"
 list ( FIND CMAKE_BUILD_RPATH ${CMAKE_BINARY_DIR} _idx )
 if ( _idx LESS 0 )
-    # not yet included in CMAKE_INSTALL_RPATH
+    # not yet included in CMAKE_BUILD_RPATH
```
haampie: Notice that CMake always sets rpaths during the build, so that just-built executables can be run. I might be missing something, but I think this bit may be redundant.

It's on install that they are cleared by default.

mmuetzel: Checking the commit history, the CMAKE_BUILD_RPATH was set since the very first version of `SuiteSparsePolicy.cmake` (and I never thought about it).

But iiuc, you are right: It should not be necessary to set it.

For a test, I've just removed it. Let's see if something breaks without it in CI.

haampie: Could be that it's CMake < 3 remnants, or something

mmuetzel: The CI passed all tests without issues. 👍

DrTimothyAldenDavis:
I can't recall.  It might date back to when cmake was first used in Mongoose or GraphBLAS.  I'm happy to remove it.

PR #602, opened 2023-12-19
GraphBLAS: Generate GB_JITpackage.c only if included files have changed.

mmuetzel:
This avoids relinking libgraphblas if nothing actually changed.
And in turn that avoids re-linking everything when calling `make demos` for LAGraph.

This might help to avoid the linker being stuck sometimes when relinking the LAGraph binaries on macOS (for an unknown reason).

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -164,8 +164,8 @@ jobs:
 
       - name: check_LAGraph
         run: |
-          cd ${GITHUB_WORKSPACE}/LAGraph/build
-          ctest .
+          cd ${GITHUB_WORKSPACE}/LAGraph
+          make demos
 
       - name: check_ParU
         run: |
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -35,7 +35,7 @@ jobs:
         # re-installed.
         run: |
           brew update
-          brew install --overwrite python@3.10 python@3.11
+          brew install --overwrite python@3.10 python@3.11 python@3.12
           brew reinstall gcc
           brew install autoconf automake ccache cmake gmp lapack libomp mpfr openblas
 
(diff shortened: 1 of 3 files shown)


mmuetzel:
The macos runner failed with the following error while installing the dependencies from homebrew:
```
==> sbt
You can use $SBT_OPTS to pass additional JVM options to sbt.
Project specific options should be placed in .sbtopts in the root of your project.
Global settings should be placed in /usr/local/etc/sbtopts

The installation does not include `sbtn`.
Error: Process completed with exit code 1.
```

I don't know what that means.

mmuetzel:
The actual error might be this: 
```
==> Pouring python@3.12--3.12.1.monterey.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/2to3-3.12
Target /usr/local/bin/2to3-3.12
already exists. You may want to remove it:
  rm '/usr/local/bin/2to3-3.12'

To force the link and overwrite all conflicting files:
  brew link --overwrite python@3.12

To list all files that would be deleted:
  brew link --overwrite --dry-run python@3.12

Possible conflicting files are:
/usr/local/bin/2to3-3.12 -> /Library/Frameworks/Python.framework/Versions/3.12/bin/2to3-3.12
/usr/local/bin/idle3.12 -> /Library/Frameworks/Python.framework/Versions/3.12/bin/idle3.12
/usr/local/bin/pydoc3.12 -> /Library/Frameworks/Python.framework/Versions/3.12/bin/pydoc3.12
/usr/local/bin/python3.12 -> /Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12
/usr/local/bin/python3.12-config -> /Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12-config
```

Edit: Should be fixed by forcing the installation of Python 3.12

DrTimothyAldenDavis:
Yes, this would be better.  I just couldn't figure out how to get it to work.

Issue #605, opened 2023-12-20
MacOS:  root CMakeLists.txt failing (not finding OpenMP_C)


DrTimothyAldenDavis:
@mmuetzel  :  on my Intel Mac, I'm getting errors when trying to build with the root CMakeLists.txt file.

It only occurs in some packages.  It occurs when trying to link a package with SuiteSparse_config, and I get reports like this:

```
CMake Error at Mongoose/CMakeLists.txt:222 (target_link_libraries):
  Error evaluating generator expression:

    $<TARGET_PROPERTY:OpenMP::OpenMP_C,INTERFACE_INCLUDE_DIRECTORIES>

  Target "OpenMP::OpenMP_C" not found.

...
CMake Error at SPEX/CMakeLists.txt:149 (target_include_directories):
  Error evaluating generator expression:

    $<TARGET_PROPERTY:OpenMP::OpenMP_C,INTERFACE_INCLUDE_DIRECTORIES>

  Target "OpenMP::OpenMP_C" not found.


CMake Error at SPEX/CMakeLists.txt:154 (target_link_libraries):
  Error evaluating generator expression:

    $<TARGET_PROPERTY:OpenMP::OpenMP_C,INTERFACE_INCLUDE_DIRECTORIES>

  Target "OpenMP::OpenMP_C" not found.


CMake Error:
  Error evaluating generator expression:

    $<TARGET_PROPERTY:OpenMP::OpenMP_C,INTERFACE_INCLUDE_DIRECTORIES>

  Target "OpenMP::OpenMP_C" not found.


CMake Error at SPEX/CMakeLists.txt:154 (target_link_libraries):
  Error evaluating generator expression:

    $<TARGET_PROPERTY:OpenMP::OpenMP_C,INTERFACE_INCLUDE_DIRECTORIES>

  Target "OpenMP::OpenMP_C" not found.
```

This occurs even though OpenMP as been found.  I don't get errors in GraphBLAS, LAGraph, ParU, or CHOLMOD, just these packages:

```
mac $ ack 'Error at' typescript | uniq -c
   4 CMake Error at Mongoose/CMakeLists.txt:222 (target_link_libraries):
   4 CMake Error at Mongoose/CMakeLists.txt:248 (target_include_directories):
   4 CMake Error at Mongoose/CMakeLists.txt:262 (target_link_libraries):
   2 CMake Error at Mongoose/CMakeLists.txt:272 (target_link_libraries):
[...]

PR #608, opened 2023-12-20
CMake: Import OpenMP targets to global scope

mmuetzel:
`SUITESPARSE_USE_OPENMP` could be set to `OFF`, but `CHOLMOD_USE_OPENMP` could be set to `ON`. So, importing that target in `SuiteSparsePolicy` only dependent on `SUITESPARSE_USE_OPENMP` is not enough in general.
Additionally, it might be cleaner to import those targets only where they are needed (even if that might be almost everywhere).

Another use-case where the current approach might lead to issues is when a downstream project imports the root CMakeLists.txt and tries to use one of its targets (that depends on OpenMP).

Importing the OpenMP targets to a global scope might help in all of those cases.

I don't know if the issue described in #605 showed up in the CI. (That seems to be passing in my fork.)
@DrTimothyAldenDavis: Does this also fix the issue you described in #605?


DrTimothyAldenDavis:
Thanks!  I'm away from the keyboard for a while but should be able to give it a try later this afternoon or tonight (Texas time I mean).

DrTimothyAldenDavis:
Looks great -- yes, this works on my Intel Mac with OpenMP from spack.  I'll try it on my M1 Mac shortly.

PR #618, opened 2023-12-22
consistent use of _USE_ and _HAS_ cmake variables

DrTimothyAldenDavis:
Variables in cmake called `*_USE_*` are always user-defined, and are not modified by any cmake script (except for GRAPHBLAS_USE_CUDA which is not yet available in production).

Variables in cmake called `*_HAS_*` are set by cmake, not by the user who builds SuiteSparse.  For example, if OpenMP is requested, the `_USE_` is `ON`/`OFF`.  Then if it is found and OK to use (v2.0 for most packages, v4.5 for ParU), then the corresponding `_HAS_` variable is set `ON`/`OFF`.


DrTimothyAldenDavis:
See issue https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/599 .

PR #621, opened 2023-12-22
add SUITESPARSE_USE_STRICT

DrTimothyAldenDavis:
* `SUITESPARSE_USE_STRICT`:

  SuiteSparse has many user-definable settings of the form `SUITESPARSE_USE_*`
  or `(package)_USE_*` for some particular package.  In general, these settings
  are not strict.  For example, if `SUITESPARSE_USE_OPENMP` is `ON` then OpenMP
  is preferred, but SuiteSparse can be used without OpenMP so no error is
  generated if OpenMP is not found.  However, if `SUITESPARSE_USE_STRICT` is
  `ON` then all `*_USE_*` settings are treated strictly and an error occurs
  if any are set to `ON` but the corresponding package or setting is not
  available.  The `*_USE_SYSTEM_*` settings are always treated as strict.
  Default: `OFF`.


DrTimothyAldenDavis:
Addresses the final part of issue https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/599

CMakeLists.txt
```diff
@@ -123,7 +123,7 @@ endif ( )
 
 if ( SUITESPARSE_USE_SYSTEM_BTF )
     list ( REMOVE_ITEM SUITESPARSE_ENABLE_PROJECTS "btf" )
-    find_package ( BTF REQUIRED )
+    find_package ( BTF 2.3.0 REQUIRED )
```
mmuetzel, line 126: Does KLU really need the exact version 2.3.0 of BTF?

The intention of the `SUITESPARSE_USE_SYSTEM_*` flags was to allow users to link to an installed version of the respective library. That might also include support for partial updates.

If that doesn't work because the user installed a library with an incompatible API, that would be a user error.
I don't know if the build system should be so restrictive (to prevent any potential user error?).

mmuetzel: Side-note: It took me a while to notice this change. It is completely unrelated to the remainder of this commit. And there is no mention of this change in the commit message...

DrTimothyAldenDavis: It should be 2.3.0 and later.  That's what the find_package parameter implies, if my understanding is correct. Using an old BYF will get the old build system and it will likely break, so I thought it would be safest to require at least the latest packages.

DrTimothyAldenDavis:
All of the packages in SuiteSparse have a version number required when they find another package in SuiteSparse itself.  For example, the KLU/CMakeLists.txt file has this:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/0940d41fae06aec9bb98024853d667c391d9f610/KLU/CMakeLists.txt#L64-L68 

This change just makes it consistent:  if compiling KLU 2.3.0 from scratch, BTF 2.3.0 (or later) is required, whether it is also being built in SuiteSparse 7.4.0, or if it is already installed in a system location.

PR #624, opened 2023-12-23
CI: Display CPU information for runner at start of job.

mmuetzel:
Show some CPU information at the start of each CI job.
(That additional information might help to find out why some runners intermittently get stuck.)


DrTimothyAldenDavis:
Great!  I'll merge it in.

PR #625, opened 2023-12-23
Set INSTALL_RPATH to $ORIGIN for all installed binaries

mmuetzel:
That is needed so that the libraries "see" each other if they are installed at a path that is not in the default LIBRARY_PATH. Apparently, `CMAKE_INSTALL_RPATH_USE_LINK_PATH` only adds RPATHs to libraries that are *not* installed in the same folder as the library itself.

See https://gitlab.com/petsc/petsc/-/merge_requests/7104#note_1706457216

Also, remove setting `LD_LIBRARY_PATH` when running the example to check if the RPATH is set correctly for the libraries.


mmuetzel:
@haampie: Would this change be ok? It adds `$ORIGIN` to the RPATH for all installed libraries. Would that be ok for your use-case?

PR #628, opened 2023-12-23
CMake: Use `@*_HAS_OPENMP@` in conditions in import target files

mmuetzel:
That was probably an oversight when the `*_USE_OPENMP` and `*_HAS_OPENMP` variables were split.


DrTimothyAldenDavis:
Oops -- yes, that's what I meant.

PR #636, opened 2023-12-25
[ParU] Mark several parameter pointers as const

gruenich:


gruenich:
Probably not as useful as I initially thought.

PR #637, opened 2023-12-25
readme for openmp; remove "set ( CMAKE_MACOSX_RPATH TRUE )" from all cmake scripts

DrTimothyAldenDavis:
`CMAKE_MACOSX_RPATH` defaults to `ON` for cmake 3.0 and later, so this is no longer needed in any cmake script of SuiteSparse (MacPorts prefers to set this `OFF`), all of which require at least 3.13 (CSparse), or 3.20 or higher (the rest of SuiteSparse).  Note that there are some cmake scripts in dependent packages with 2.x requirements, but those cmake scripts are not used to build those packages; SuiteSparse uses its own cmake scripts to build them.

Also added more discussion of `*_USE_OPENMP` cmake options in the top level README.

See https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_RPATH.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -811,34 +811,42 @@ build type).  The static libraries will not be built (since
 
   If `ON`, OpenMP is used if it is available.  Default: `OFF`.
 
-  UMFPACK, CHOLMOD, SPQR, and GraphBLAS will be slow if OpenMP is not used.
+  GraphBLAS, LAGraph, and ParU will be slow if OpenMP is not used.
+  Other packages (UMFPACK, CHOLMOD, SPQR) may use OpenMP in the BLAS/LAPACK,
+  which is essential for performance.
 
   Note that BLAS and LAPACK may still use OpenMP internally; if you wish to
   disable OpenMP in an entire application, select a single-threaded
   BLAS/LAPACK.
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -811,34 +811,42 @@ build type).  The static libraries will not be built (since
 
   If `ON`, OpenMP is used if it is available.  Default: `OFF`.
 
-  UMFPACK, CHOLMOD, SPQR, and GraphBLAS will be slow if OpenMP is not used.
+  GraphBLAS, LAGraph, and ParU will be slow if OpenMP is not used.
```
mmuetzel, line 814: IIUC, `CHOLMOD_CUDA` and the Supernodal module of CHOLMOD also use OpenMP (directly - not just via an accelerated BLAS or LAPACK implementation).

DrTimothyAldenDavis: Yes, but it's modest impact to the performance of CHOLMOD (either the GPU or CPU).  CHOLMOD gets most of its parallel performance inside the BLAS/LAPACK.

I will update the readme to clarify this point.

DrTimothyAldenDavis: see https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/b9a18492097424e169f5ec786799d609af7ab006

PR #658, opened 2023-12-30
CMake: Find only required OpenMP components.

mmuetzel:
That speeds up configuration and avoids bogus warnings and errors if, e.g., the installed Fortran toolchain doesn't support OpenMP.


DrTimothyAldenDavis:
Looks great.  I can add this, and the others, to 7.5.0.beta1.

PR #659, opened 2023-12-30
Example: Remove explicit import of transient dependencies.

mmuetzel:
That should no longer be needed. And removing them should allow the CI to check if the CMake Config files are complete.

### DIFF
--- a/Example/CMakeLists.txt
+++ b/Example/CMakeLists.txt
@@ -107,14 +107,6 @@ find_package ( SPEX 2.3.0 REQUIRED )    # requires GMP and MPFR
 find_package ( SPQR 4.3.0 REQUIRED )
 find_package ( UMFPACK 6.3.0 REQUIRED )
 
-# look for all external libaries:
-find_package ( OpenMP REQUIRED )
-find_package ( BLAS REQUIRED )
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Looks great -- thanks.

DrTimothyAldenDavis:
I'm seeing errors when building the Example when not using the top level CMakeLists.txt.  After "make local ; make install ; cd Example ; make" I get:

```
CMake Error at /home/davis/dev2/SuiteSparse/lib/cmake/ParU/ParUTargets.cmake:62 (set_target_properties):
  The link interface of target "SuiteSparse::ParU_static" contains:

    OpenMP::OpenMP_CXX

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

  /home/davis/dev2/SuiteSparse/lib/cmake/ParU/ParUConfig.cmake:121 (include)
  CMakeLists.txt:104 (find_package)
```

mmuetzel:
Looking into it.

mmuetzel:
PR #662 should fix this issue.

PR #662, opened 2023-12-30
ParU: Automatically import OpenMP when importing CMake target

mmuetzel:
Untested. Let's see what the CI says.


mmuetzel:
Looks good in CI.

Issue #656, opened 2023-12-30
Static linker flags wrong in LAGraph.pc


mmuetzel:
The static linker flags are wrong in the pkg-config file generated by LAGraph when building with the root Makefile (or as a stand-alone project). They contain something that looks like a CMake target name of GraphBLAS verbatim.
The flags in that file look ok when the file is generated with the root CMakeLists.txt.

I'm looking into the cause and a possible fix currently.

mmuetzel:
Looking at the pkg-config files in more detail revealed a couple more issues.
I opened some PRs that should hopefully fix them: #660, #661

I'll wait if they get accepted before looking into adding CI tests for them.

mmuetzel:
I opened #666 that adds steps to the CI which might have caught those issues earlier.

DrTimothyAldenDavis:
I've posted a 7.5.0.beta3 with this fix: https://github.com/DrTimothyAldenDavis/SuiteSparse/releases/tag/v7.5.0.beta3 .

Are these issues fixed now?

mmuetzel:
The CI step that builds the example using autotools isn't failing. It would have failed if this still was an issue.

So, this particular issue should be fixed now. 👍

PR #666, opened 2023-12-31
Add build files using autotools to Example package

mmuetzel:
This adds files to the Example package that lay out how downstream projects that use autotools could integrate SuiteSparse.
This leaves the CMake build rules of the the example (mostly) untouched.

Additionally, this adds a step to the CI which builds the example using autotools. This can serve as a test of whether the .pc files are working. (This test would probably have revealed issues like #656 earlier.)

We already knew that flags were missing from the .pc files to statically link with CUDA. I still don't know enough about CUDA to determine the necessary linker flags generally. So, I resorted to hard-coding what I found on their forum. I hope that is good enough for now.

I chose autotools over other build systems (like meson, scons, ...) because I know how to use autotools but not the others. Additionally, autotools (albeit being "old") is still one of the major established build systems. 
If others would like to contribute examples that show how to integrate SuiteSparse in other build systems, that might also be nice.


DrTimothyAldenDavis:
Julia is having trouble: https://github.com/JuliaLang/julia/pull/52577#issuecomment-1872948735

PR #675, opened 2024-01-01
Example: Add tests that use each sub-project separately

mmuetzel:
This checks if each `*Config.make` file is really self-contained. 
Importing all targets (for the existing test, i.e., building the Example package) was hiding an issue in the `KLUConfig.cmake` file (#664). The new tests would probably have revealed that issue earlier.

These tests also showed an issue in `LDLConfig.cmake`.

The code in the new examples are mainly taken from the existing example package. That means that there are no "proper" tests yet for ParU (or Mongoose).

Eventually, we could probably extend that to check if each *.pc file separately contains sufficient information, too.


DrTimothyAldenDavis:
Looks good.  Eventually I could add some more significant tests to each of these.

DrTimothyAldenDavis:
I like the idea of adding these tests, but I think it's best to keep the Example package as simple as possible.  The Example is meant primarily as an example of how to create a library that depends on one or more SuiteSparse packages.

So I've split the new tests into a new folder, SuiteSparse/TestConfig, with this commit:

https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/630812260209d55f7aa9a9fc2734ba7700521feb

I also found and fixed some errors in the colamd and ccolamd tests, and a few memory leaks in other tests.

mmuetzel:
Ok...
I thought those could serve as examples for linking to one single library of SuiteSparse. But moving those examples to a different folder is also fine. They won't be lost after all.

Testing the validity of the *Config.cmake files is just a "secondary usage" of those small example programs imho.

PR #678, opened 2024-01-02
Fix typos

mmuetzel:


GraphBLAS/lz4/lz4hc.h
```diff
@@ -305,7 +305,7 @@ LZ4LIB_API void LZ4_resetStreamHC (LZ4_streamHC_t* streamHCPtr, int compressionL
  * They should not be linked from DLL,
  * as there is no guarantee of API stability yet.
  * Prototypes will be promoted to "stable" status
- * after successfull usage in real-life scenarios.
+ * after successful usage in real-life scenarios.
```
DrTimothyAldenDavis: I'd like to leave lz4 as-is.  The GraphBLAS/lz4 directory is an exact copy of a subset of the lz4 library.  The GraphBLAS/lz4/README.txt file has my notes about lz4, and it states:

```
This directory contains a minimal copy of lz4 v1.9.3, from
https://github.com/lz4/lz4.git and http://www.lz4.org, by Yann Collet.  See
./LICENSE and ./README.md for more details.  Four source files are used in
SuiteSparse:GraphBLAS: lz4/lib/lz4.[ch], and lz4/lib/lz4hc.[ch] copied here
unmodified from v1.9.3, which are under the BSD-2-clause license.
```

so I'd rather not change anything in this file if I don't have to, not even typos.

mmuetzel: I basically used an equivalent to `sed` to replace every "successfull" with "successful".
I can revert the change in this file if you prefer.

mmuetzel: Reverted that part

PR #687, opened 2024-01-03
ParU: Use correct format specifiers in demos

mmuetzel:
Also use C++ idioms for printing to stdout.

### DIFF
--- a/ParU/Demo/paru_demo.cpp
+++ b/ParU/Demo/paru_demo.cpp
@@ -11,7 +11,10 @@
  *
  * @author Aznaveh
  * */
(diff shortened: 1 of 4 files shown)


DrTimothyAldenDavis:
Looks fine.  Yes, ParU has lots of C-style code that could be done in C++ style instead.  I missed the branch 'dev' though.  dev2 is more up to date than dev at the moment so I just copied the files over and pushed it there: https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/cea4de9df5d19faa9e82f6c79994b421c15e416d .

Issue #682, opened 2024-01-03
[PowerPC] SuiteSparse_CHOLMOD 7.4.0 build failure: `cholmod_dbound.c: error: two or more data types in declaration specifiers` (Libm already defines Real)


barracuda156:
```
[ 43%] Building C object CMakeFiles/CHOLMOD.dir/Utility/cholmod_dbound.c.o
[...]

PR #690, opened 2024-01-03
CHOLMOD: Don't use macOS extensions of math library

mmuetzel:
Limit scope of workaround to affected libraries only.

Also use the correct name of the preprocessor definition `__NOEXTENSIONS__`. (See: https://opensource.apple.com/source/Libm/Libm-92/i386.subproj/math.h.auto.html)

Should be fixing #682.

### DIFF
--- a/CHOLMOD/CMakeLists.txt
+++ b/CHOLMOD/CMakeLists.txt
@@ -351,6 +351,13 @@ if ( BUILD_SHARED_LIBS )
     target_include_directories ( CHOLMOD
         INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/Include>
                   $<INSTALL_INTERFACE:${SUITESPARSE_INCLUDEDIR}> )
+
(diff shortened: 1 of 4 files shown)


DrTimothyAldenDavis:
Yes, that would be better.  But it also needs to be added to KLU/CMakeLists.txt too.

mmuetzel:
> But it also needs to be added to KLU/CMakeLists.txt too.

Oops. It's probably only needed in KLU...
https://github.com/macports/macports-ports/commit/e485b0a0c8125cb9d8bcc6241a9c2016514188c8#diff-93ed37ccf33fb803f40494f870c28d04b1b3cb795de482ce3c536a49c75f39b1

mmuetzel:
Grepping for `Real` and looking at where those files are included, the broader approach might have been the better one.
Apart from KLU and CHOLMOD, it's also `umf_version.h` from UMFPACK (and hence by extension also ParU) and Tcov in SPQR (and maybe others?) that need that change...

PR #693, opened 2024-01-03
Disable extensions in math.h with older versions of macOS for all projects

mmuetzel:
Afaict, more libraries apart from KLU and CHOLMOD might need this workaround on older versions of macOS.

This essentially reverts to your original change (with the typo fixed). 

Sorry for the confusion.


DrTimothyAldenDavis:
Sure thing.   This is probably a good idea since I don't think any of the packages in SuiteSparse need these extensions.

Revising the fix for https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/682

DrTimothyAldenDavis:
And in the future, when I revised UMFPACK to handle single precision matrices like CHOLMOD, I might use the same thing, as

`#define Real float
`
or 

`#define Real double
`

PR #699, opened 2024-01-04
UMFPACK: Check syntax for novector pragma and use it

mmuetzel:
Same as #688 but for the `novector` pragma.


DrTimothyAldenDavis:
Looks fine.  It's a lot of work to keep a workaround for a very old Intel compiler though.  I might just delete the novector pragma entirely when I take a pass at UMFPACK.

PR #700, opened 2024-01-04
SPEX: Silence compiler warnings about unknown pragmas

mmuetzel:
Compilers that don't recognize `#pragma GCC` (e.g., MSVC) emit warnings when they see those pragmas.
Add a condition (for GCC compatible compilers like gcc, clang, icc, ...) to silence those warnings for non-GCC compatible compilers.


mmuetzel:
Added a similar change to the sources for the GraphBLAS compressed JIT source file generator.

PR #702, opened 2024-01-05
Properly clean after ParU PDF user guide compilation

svillemot:
For consistency with other PDFs in SuiteSparse.

I’m sorry but I won’t sign the CLA (for philosophical and practical reasons – I would have to involve my employer).

However this contribution is so small that it is likely not copyrightable. So I don’t think signing the CLA would make a difference here.


DrTimothyAldenDavis:
I would normally be required to get a CLA from the constraints of my organization, myself.  But you're right, this seems fine to me at least, without one.

PR #704, opened 2024-01-05
SPQR: Slightly overhaul code of GPU libraries

mmuetzel:
This mostly avoids warnings with some compilers. If I understand those warnings correctly, they point to valid "issues":

* Incorrect size of format specifiers when printing to stdout. Avoid that by using a C++ style output stream.
* Missing placement delete operator for overloaded placement new operator. Avoid that issue by using the placement new/delete operators from the STL instead.
* Incomplete template class definition for explicit instantiation of the template class. Avoid that issue by defining the entire template class in a single compilation unit.

### DIFF
--- a/SPQR/GPUQREngine/CMakeLists.txt
+++ b/SPQR/GPUQREngine/CMakeLists.txt
@@ -66,7 +66,12 @@ configure_file ( "Config/GPUQREngine.hpp.in"
 
 #-------------------------------------------------------------------------------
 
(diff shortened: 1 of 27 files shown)


DrTimothyAldenDavis:
It's been a while since I touched this code, and also a long time since I've worked with the C++ placement new technique.  It will take me some time to get back up to speed and try it out.

DrTimothyAldenDavis:
I'd like to converge to a stable SuiteSparse 7.5.0 release soon.  I'll try to include these changes.

DrTimothyAldenDavis:
My GPU tests are passing so I'll merge this in.

PR #705, opened 2024-01-05
CHOLMOD: Call function in GPU code

mmuetzel:
To be honest, I don't understand that part of the code. But having a sole statement with the function address doesn't look correct either.


DrTimothyAldenDavis:
oops.  Yes, that's a mistake.

PR #707, opened 2024-01-05
GraphBLAS: Silence compiler warnings about values outside `GrB_Field`

mmuetzel:
Cast the switch argument to `int` to avoid compiler warnings (with clang) because some of the case values are not in `GrB_Field` but are part of different enumerations.


DrTimothyAldenDavis:
This looks familiar ... but only because I fixed it on my M1 Mac just a few days ago but forgot to push the changes.  They're still there, and my changes are identical to yours.  So I'll go ahead and merge your changes in and checkout mine on my M1.

PR #717, opened 2024-01-09
CHOLMOD: Use object libraries to simplify source.

mmuetzel:
Many files in CHOLMOD are built multiple times with different preprocessor macros. This is currently done with stub files that define a macro and then include a file with the actual code.

Avoid using those stub files by compiling the same source file multiple times with different flags using CMake object libraries.

I don't know if this is something that is acceptable. I haven't done many checks with these changes. So, keeping as draft for now.


DrTimothyAldenDavis:
I'd prefer to leave the setup as-is, so that each file is compiled just once.  It works, I understand it, and changing it would cause a lot of changes in all of SuiteSparse (I use this strategy in many packages).  I'm concerned that I would end up with build system that I might not be able to revise or understand.

DrTimothyAldenDavis:
This change would also break my CHOLMOD/Tcov tests, which doesn't use cmake at all and requires these files.

PR #728, opened 2024-01-12
SuiteSparse 7.5.1

DrTimothyAldenDavis:


DrTimothyAldenDavis:
I'll add a test for these new version macros, as well.

PR #730, opened 2024-01-12
GraphBLAS: Disable JIT compiler by default when cross-compiling.

mmuetzel:
I was hoping that users that cross-compile libraries to a different target understand their toolchains. But that doesn't seem to be the case generally.
To improve the out-of-the-box experience for those users, disable the JIT compiler by default when cross-compiling. Interested users could still enable it (and supply settings for a working native toolchain).


DrTimothyAldenDavis:
The JIT is extremely important for performance.  Cross-compiling is something not done by a typical user, but by someone who should know how to use toolchains on both the host and the target.  I'd rather have them try to get the JIT to work, and in the worst case, if they can't get it to work then they can select a non-default option to turn off the JIT.

mmuetzel:
That sounds reasonable. We probably just have to live with users that come with unreasonable expectations.

PR #737, opened 2024-01-16
Correction description

DragonPara:
max_udiag is the maxinum abs value on diagonal of D


DrTimothyAldenDavis:
Thanks for catching it.  There's a typo in your update, though.  It should be "maximum".  Also, did you recently send me an email with a signed contributor agreement?  If so, that agreement doesn't list your name in an form that I can read.  Could you resend it?  (Thanks).  I realize this is a tiny change and thus likely has no bearing on copyright or licensing, but my organization requires me to ask for a contributor agreement ... sorry for the hassle.

DragonPara:
Hi doctor, I just sent you a signed contributor agreement and corrected my typo error.

PR #739, opened 2024-01-16
CHOLMOD: Fix typo in build rules.

mmuetzel:
`CMAKE_EXTRA_INCLUDE_FILES` is currently not reset to its previous value as it was intended.
Even if this typo doesn't have actual consequences currently afaict, it should still be fixed.


DrTimothyAldenDavis:
I'll include this in CHOLMOD 5.1.2, which I'd like to release soon, to fix this bug: https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/741

PR #743, opened 2024-01-19
SuiteSparse 7.6.0

DrTimothyAldenDavis:
various minor bug fixes:

```
remove `set ( package_USE_OPENMP OFF )` if `SUITESPARSE_USE_OPENMP` is set OFF.
resolve various compiler warnings
fix ABI compatibility for CHOLMOD
CI updates for additional testing
```


mmuetzel:
Even if there aren't new features in that version, should it be called SuiteSparse 7.6.0 because we know there will be an ABI breaking change with respect to 7.5.1?

DrTimothyAldenDavis:
I don't have a strong preference either way.  @svillemot : what do you think about the version number for the ABI fix to CHOLMOD?  CHOLMOD would go from 5.1.1 to 5.1.2 (an minor bug fix to repair the ABI problem), and then SuiteSparse itself would go from 7.5.1 to 7.5.2 or (as @mmuetzel suggests) to 7.6.0, to reflect the change to CHOLMOD.

And preference?

svillemot:
I don’t have a preference for the version number of CHOLMOD and SuiteSparse, as long as the SOVERSION is increased.

svillemot:
I don’t have a preference for the version number of CHOLMOD and SuiteSparse. I guess it should be treated as any bugfix (unless you plan to introduce more changes in the next release).

DrTimothyAldenDavis:
I'd prefer to use SuiteSparse 7.5.2 (where 7 is the main version and also the SO version), and CHOLMOD is 5.1.2 (5 being both the main version and SO version).

I view this as a serious bug fix so CHOLMOD 5.1.1 -> 5.1.2 seems fine to me, and thus SuiteSparse 7.5.1 -> 7.5.2 is my preference.

I realize the main version and SOVERSION don't have to match, but I would find that confusing to maintain.  SuiteSparse has lots of packages, and I already have to keep track of 3 numbers x.y.z for each package.  I don't want to add more numbers, so I prefer CHOLMOD 5.x.x to compile to libcholmod.so.5.

Does this sound OK?

If this is OK, I can release a stable SuiteSparse 7.5.2 today.

svillemot:
Sounds good to me, thanks.

mmuetzel:
If it is not too late, could you still consider #745 that would fix an issue for users which use a toolchain without OpenMP?

Independent of the SOVERSION, I was assuming that the major and minor version would indicate that libraries are ABI compatible. Like you wrote, the third version number would be used for non-ABI breaking bug fixes.
This bug fix breaks the ABI (with respect to, e.g., SuiteSparse 7.5.1). So, would it be better to bump the minor version for it?

ChangeLog
```diff
@@ -1,3 +1,33 @@
+Jan 20, 2024: version 7.5.2
+
+    * CHOLMOD 5.1.2: bug fix (restore ABI compatibility with 5.0.x)
```
mmuetzel: Could it be made clearer that this is an ABI-breaking change with respect to CHOLMOD 5.1.X?
``` suggestion
    * CHOLMOD 5.1.2: bug fix (restore ABI compatibility with 5.0.x, i.e., 5.1.2 is ABI incompatible to 5.1.x)
```
This sound self-contradicting. But that sentence would make more sense if the upcoming version would be CHOLMOD 5.2.0 instead.

CHOLMOD/Doc/ChangeLog
```diff
@@ -1,3 +1,11 @@
+Jan 20, 2024: version 5.1.2
+
+    * bug fix:  CHOLMOD 5.1.0 added a new variable in the CHOLMOD Common
+        struct for development only but this change broke ABI compatibility
+        since the SOVERSION was not bumped at the same time.  v5.1.2 reverts
+        this change so the new variable does not appear in production, thus
+        preserving ABI compatibility with v5.0.x.
```
mmuetzel:
``` suggestion
        restoring ABI compatibility with v5.0.x.
```

ChangeLog
DrTimothyAldenDavis: Yes -- that does make more sense.  I'll do CHOLMOD 5.2.0 and SuiteSparse 7.6.0.

DrTimothyAldenDavis:
I've revised the version numbers of CHOLMOD to 5.2.0, and SuiteSparse to 7.6.0.  These changes propagate to CHOLMOD's dependencies (all packages in SuiteSparse that depend on CHOLMOD should look for CHOLMOD 5.2.0, not an earlier version, to make sure they all use a consistent ABI for CHOLMOD).

DrTimothyAldenDavis:
> If it is not too late, could you still consider #745 that would fix an issue for users which use a toolchain without OpenMP?

I've merged #745 for the OpenMP issue.  @mmuetzel : Does that change resolve it, at least for SuiteSparse 7.6.0?

@svillemot :  I think this PR will become the stable SuiteSparse 7.6.0.  I will release a SuiteSparse 7.6.0.beta1 soon with these changes.

PR #746, opened 2024-01-23
Windows compatibility

DragonPara:
Use the comment pound key in Makefile as the first column and set ';' Change to '&&' to increase compatibility
When I compiled it with makefile and Visual Studio 2022, I found it.


DrTimothyAldenDavis:
Thanks; I'll include this in SuiteSparse 7.6.0.

Issue #753, opened 2024-02-02
GraphBLAS uses `_Generic` selections with potentially compatible types


mmuetzel:
**Describe the bug**
When trying to build LAGraph with `clang-cl`, it fails with errors like the following:
```
D:\repo\SuiteSparse\SuiteSparse\LAGraph\src\utility\LAGr_Init.c(195,18): error: type 'GxB_Option_Field' in generic association compatible with previously specified type 'int'
        GRB_TRY (GxB_set (GxB_NTHREADS, LG_nthreads_inner)) ;
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
D:\repo\SuiteSparse\SuiteSparse\GraphBLAS\Include\GraphBLAS.h(3969,13): note: expanded from macro 'GxB_set'
            GxB_Option_Field : GxB_Global_Option_set ,          \
            ^~~~~~~~~~~~~~~~
D:\repo\SuiteSparse\SuiteSparse\LAGraph\include\LAGraph.h(366,28): note: expanded from macro 'GRB_TRY'
    GrB_Info LG_GrB_Info = GrB_method ;         \
                           ^~~~~~~~~~
D:\repo\SuiteSparse\SuiteSparse\LAGraph\src\utility\LAGr_Init.c(195,18): note: compatible type 'int' specified here
        GRB_TRY (GxB_set (GxB_NTHREADS, LG_nthreads_inner)) ;
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
D:\repo\SuiteSparse\SuiteSparse\GraphBLAS\Include\GraphBLAS.h(3968,13): note: expanded from macro 'GxB_set'
            int              : GxB_Global_Option_set ,          \
            ^~~
D:\repo\SuiteSparse\SuiteSparse\LAGraph\include\LAGraph.h(366,28): note: expanded from macro 'GRB_TRY'
    GrB_Info LG_GrB_Info = GrB_method ;         \
                           ^~~~~~~~~~
1 error generated.
```

**To Reproduce**
Configure with the following using a MSVC build environment:
```
[...]

PR #755, opened 2024-02-03
GraphBLAS: Avoid possibility of compatible types in `_Generic` selections

mmuetzel:
If I understand the motivation for these generic selections correctly, they are meant to provide "specializations" if the input is part of some enumerations. The "generic" function is given by the line with the `int` type.
The `int` type in those lists can cause issues if `int` is the underlying type for enumerations (in some implementations).

Avoid that possible issue by using the keyword `default` for the "generic" function.

This should be fixing #753.

Additionally, avoid adding incompatible command line options for `clang-cl` in the build rules of LAGraph (a clang compiler with MSVC-compatible command line interface).


DrTimothyAldenDavis:
Great solution; I was worried that I would have to introduce a major breaking change to fix that macro, forcing me to go to GarphBLAS 10.0.0.  This change doesn't need that.

PR #761, opened 2024-02-06
CXSparse: Don't use complex numbers for MSVC targets

mmuetzel:
For `clang`/`clang++` targeting MSVC, the CMake variable `MSVC` is not set (only for `clang-cl` targeting MSVC).
However, the MSVC target uses different types for complex numbers that CXSparse doesn't support (yet?).

Deactivate complex number arithmetic for CXSparse also for any compiler that targets MSVC.

Also, avoid overlinking the CXSparse library with libsuitesparseconfig. (It only uses the header but no functions from it.)

### DIFF
--- a/CXSparse/CMakeLists.txt
+++ b/CXSparse/CMakeLists.txt
@@ -45,7 +45,7 @@ include ( SuiteSparsePolicy )
 # MS Visual Studio does not support the complex type for ANSI C11.
 # FIXME: see GraphBLAS for how to use complex types in MS Visual Studio.
 
-if ( MSVC )
+if ( MSVC OR ("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC") )
     option ( CXSPARSE_USE_COMPLEX "ON: complex data type enabled.  OFF (default): complex data type disabled." OFF )
 else ( )
     option ( CXSPARSE_USE_COMPLEX "ON (default): complex data type enabled.  OFF: complex data type disabled." ON )


DrTimothyAldenDavis:
Yes, CXSparse doesn't support MSVC's half-baked complex type.  I have tagged it as a possible future extension:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/9706d509e4d3058653247b1030da46a16f2ea41e/CXSparse/CMakeLists.txt#L45-L46

Regarding overlinking:  I'd like to revise CXSparse/Source/cs_malloc so that it calls SuiteSparse_config_malloc and friends, which themselves call malloc by default.  I just haven't gotten around to it yet.  So I'd prefer to keep the overlinking for now, or at most just comment it out so I can easily add it back in.

mmuetzel:
> Regarding overlinking: I'd like to revise CXSparse/Source/cs_malloc so that it calls SuiteSparse_config_malloc and friends, which themselves call malloc by default. I just haven't gotten around to it yet. So I'd prefer to keep the overlinking for now, or at most just comment it out so I can easily add it back in.

Oops. I now remember that you wrote the exact thing a couple of weeks ago. Sorry, I forgot.

Stripped that part of the PR.

DrTimothyAldenDavis:
I think this is a good reminder for me to revise CXSparse to use the SuiteSparse_config_malloc functions and friends anyway.  I'll do that shortly.

DrTimothyAldenDavis:
In particular, see : https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/3e43bffa3ab9323eb886a0484071482e75ac35df

PR #766, opened 2024-02-20
Fix spex

whuaegeanse:
Revise the version AMD in SPEX


DrTimothyAldenDavis:
Oops. Thanks for catching that.

PR #775, opened 2024-03-08
SPEX User guide update

clouren:
Updating the SPEX user guide with new macros and gmp/mpq/mpfr functions


DrTimothyAldenDavis:
I revised Config/SPEX.h.in to match the changes to SPEX.h (removing the FIXME comments).

PR #787, opened 2024-03-22
CI: Add a runner (MinGW/UCRT64) that builds without Fortran compiler.

mmuetzel:
The CLANG32 (that is now failing and has been removed) checked if SuiteSparse could be built successfully in the absence of a Fortran compiler.

This PR adds a new runner that installs a build environment where the Fortran compiler is omitted (on purpose).

The UCRT64 environment of MSYS2 uses a GNU toolchain (GCC, GNU ld, libstdc++, ...) that links against the (newer) C runtime of Windows called UCRT (Universal CRT). The latter is the main difference to the MINGW64 environment which links against the (traditional) MSVCRT. 
See: https://www.msys2.org/docs/environments/

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -309,7 +309,7 @@ jobs:
       matrix:
 #       CLANG32 disabled for now:
 #       msystem: [MINGW64, MINGW32, CLANG64, CLANG32]
-        msystem: [MINGW64, MINGW32, CLANG64]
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
So this test should bypass the failing CLANG32, while at the same time testing the no-Fortran case, right?

DrTimothyAldenDavis:
If that's my understanding then that's a good idea to add -- thanks.

PR #793, opened 2024-04-01
Fix a find_package(UMFPACK) bug in ParU/CMakeLists.txt

valgur:
External `UMFPACK` is currently not being found due to a small bug in `ParU/CMakeLists.txt`.

### DIFF
--- a/ParU/CMakeLists.txt
+++ b/ParU/CMakeLists.txt
@@ -111,7 +111,7 @@ if ( NOT SUITESPARSE_ROOT_CMAKELISTS )
 
     find_package ( UMFPACK 6.3.3
         PATHS ${CMAKE_SOURCE_DIR}/../UMFPACK/build NO_DEFAULT_PATH )
-    if ( NOT CAMD_FOUND )
+    if ( NOT UMFPACK_FOUND )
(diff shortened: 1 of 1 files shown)


mmuetzel:
Oof. That's probably a copy-paste error ~~that wasn't caught by CI because `CAMD_FOUND` happened to be set after `find_package(CHOLMOD)`~~.

Your change looks good to me. Thank you for your contribution. 👍

Edit: Strike that previous remark. The CI didn't catch it because it is linking to the CHOLMOD libraries in the *common* build or source tree. 🤦
Your change still looks good though. 👍

DrTimothyAldenDavis:
Thanks for the catch.

I realize this is a very minor change, but my organization requires that any contributor signs a contributor agreement (a copy is at the top-level in SuiteSparse).  Can you sign it and send it to me?

mmuetzel:
This change fixes a bug in the current CMake rules. I'm assuming that @valgur didn't respond in more than 6 weeks. With that in mind, what should be done with this change?

valgur:
I emailed my signed contributor license now. Apologies for the delay.

PR #795, opened 2024-04-17
CHOLMOD: Add preprocessor flags that affect cholmod.h to targets and .pc

### LINKED ISSUE #794: Way to detect whether CHOLMOD is compiled with LGPL enabled
Thank you for making this large collection of very useful libraries available.

I'm offering my users the ability to link my MIT licensed code to cholmod using cmake. For this purpose it is sometimes necessary to use CHOLMOD in LGPL mode, i.e. no supernodal, cuda etc. Otherwise the combined code becomes GPL, which is in some cases too restrictive. This means I'd like to build in a safety feature in my compilation process that requires the user to explicitly and knowingly enable the GPL features of CHOLMOD, or otherwise only allow linking to the LGPL version. 

To do this in cmake I require a way to detect how CHOLMOD was compiled. However, the compile definitions are not carried over to the installed target as `add_compile_definitions(NGPL)` etc are used. Would it be possible to use something like `target_compile_definitions(CHOLMOD PUBLIC NGPL)` instead (also for NSUPERNODAL), similar to how it is currently done for `CHOLMOD_HAS_CUDA`. I believe this then should carry over to the cmake installed target. This would then automatically enable the #ifdef features in cholmod.h as well. 

Or is there perhaps another way to achieve what I need?

mmuetzel:
Some preprocessor flags that are conditionally set by the build rules of CHOLMOD affect which symbols are exported from the built libraries. 
Add the preprocessor flags that are used to guard function declarations in `cholmod.h` to the CMake targets and the pkg-config file. That way consumers of the library get only those declarations that are actually exported from the library when including `cholmod.h` in their code.

Fixes #794.

The names of these preprocessor macros are pretty generic (e.g., `NGPL`) and might collide with symbols of the same name in other projects. It might be a good idea to rename them to something more specific (e.g., `CHOLMOD_NGPL`).
But I don't know if backwards compatibility is required for those names. (Maybe better to handle a potential renaming in a different change anyway...)

### DIFF
--- a/CHOLMOD/CMakeLists.txt
+++ b/CHOLMOD/CMakeLists.txt
@@ -52,7 +52,7 @@ if ( NOT CHOLMOD_GPL )
     set ( CHOLMOD_MODIFY OFF )
     set ( CHOLMOD_SUPERNODAL OFF )
     set ( CHOLMOD_USE_CUDA "OFF" CACHE STRING "" FORCE ) # OK: disabled by user-setting of CHOLMOD_GPL
-    add_compile_definitions ( NGPL )
+    set ( CHOLMOD_CFLAGS "${CHOLMOD_CFLAGS} -DNGPL" )
 endif ( )
 
 include ( CheckTypeSize )
@@ -165,7 +165,7 @@ endif ( )
     endif ( )
     if ( NOT CHOLMOD_CHECK )
         # if CHOLMOD_CHECK is OFF: do not build the Check module
-        add_compile_definitions ( NCHECK )
+        set ( CHOLMOD_CFLAGS "${CHOLMOD_CFLAGS} -DNCHECK" )
     endif ( )
 
     #---------------------------------------------------------------------------
@@ -198,7 +198,7 @@ endif ( )
         # build the Supernodal or Modify modules that depend on it.
         set ( CHOLMOD_SUPERNODAL OFF )
         set ( CHOLMOD_MODIFY OFF )
-        add_compile_definitions ( NCHOLESKY )
+        set ( CHOLMOD_CFLAGS "${CHOLMOD_CFLAGS} -DNCHOLESKY" )
     endif ( )
 
     #---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
I'm already working on a solution

DrTimothyAldenDavis:
I would like to do something different, by adding #cmakedefine's to cholmod.h.in, which then cmake configures to create cholmod.h.

The draft is in my paru_dev2 branch.  I would have pushed it to dev2 as I usually do, but I'm in the middle of making a ParU version 1.0.0.  ParU depends on CHOLMOD.

DrTimothyAldenDavis:
My plan is to have these configured in cholmod.h:

```
// These flags are configured by cmake when CHOLMOD is compiled:
#define CHOLMOD_HAS_GPL
#define CHOLMOD_HAS_CHECK
#define CHOLMOD_HAS_CHOLESKY
#define CHOLMOD_HAS_CAMD
#define CHOLMOD_HAS_PARTITION
#define CHOLMOD_HAS_MATRIXOPS
#define CHOLMOD_HAS_MODIFY
#define CHOLMOD_HAS_SUPERNODAL
#define CHOLMOD_HAS_CUDA
#define CHOLMOD_HAS_OPENMP
```

and also to add a run-time check:
```

//------------------------------------------------------------------------------
// CHOLMOD query: check for CHOLMOD configuration
//------------------------------------------------------------------------------

typedef enum
{
    CHOLMOD_QUERY_HAS_GPL         = 0,  // has GPL licensed modules
    CHOLMOD_QUERY_HAS_CHECK       = 1,  // has Check Module
    CHOLMOD_QUERY_HAS_CHOLESKY    = 2,  // has Cholesky Module
    CHOLMOD_QUERY_HAS_CAMD        = 3,  // has CAMD and CCOLAMD
    CHOLMOD_QUERY_HAS_PARTITION   = 4,  // has Partition Module
    CHOLMOD_QUERY_HAS_MATRIXOPS   = 5,  // has MatrixOps Module
    CHOLMOD_QUERY_HAS_MODIFY      = 6,  // has Modify Module
    CHOLMOD_QUERY_HAS_SUPERNODAL  = 7,  // has Supernodal Module
    CHOLMOD_QUERY_HAS_CUDA        = 8,  // has GPU Module
    CHOLMOD_QUERY_HAS_OPENMP      = 9   // has OpenMP
}
cholmod_query_t ;

bool cholmod_query      // true if CHOLMOD has a specific feature
(
    cholmod_query_t feature
) ;
bool cholmod_l_query (cholmod_query_t feature) ;
```

PR #801, opened 2024-05-14
CI: Add riscv64 arch

suvorovrain:
CI for `riscv64` runs on alpine edge branch. `Valgrind` packages are not available on this architecture, so I had to remove it from the dependency list. The directory for saving the `ccahce` cache is different on the edge branch, so I also changed it in `riscv64` arch case. Also i had to disable static libs flag in `Cmake` examples. The problem is that the package versions in the `edge` branch are different from the `stable` one, and when i tried to statically link i got following: `lto1: fatal error: bytecode stream in file '/usr/lib/libgmp.a' generated with LTO version 13.0 instead of the expected 13.1` (on other architectures in the edge branch the problem is same)

### DIFF
--- a/.github/workflows/build-arch-emu.yaml
+++ b/.github/workflows/build-arch-emu.yaml
@@ -41,22 +41,16 @@ jobs:
             ccache-max: 80M
             extra-build-libs: ":GraphBLAS:LAGraph"
             extra-check-libs: ":GraphBLAS:LAGraph"
-            branch: latest-stable
           - arch: aarch64
             ccache-max: 42M
-            branch: latest-stable
           - arch: armv7
             ccache-max: 42M
-            branch: latest-stable
           - arch: ppc64le
             ccache-max: 45M
-            branch: latest-stable
           - arch: s390x
             ccache-max: 42M
-            branch: latest-stable
           - arch: riscv64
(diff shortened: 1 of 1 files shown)


mmuetzel:
Thank you for your PR.

IIUC, Alpine Linux aims to release a new stable version each May and November [[1]](https://alpinelinux.org/releases/).
According to their (stub) [Wiki page](https://wiki.alpinelinux.org/wiki/Riscv64), they are planning to include support for RISCV64 in 3.20 which will probably be released soon.

Maybe, we should just wait a couple of weeks for that release. That would probably make it a lot easier to add CI for that platform.

Edit: Fwiw, a first release candidate for Alpine Linux 3.20 has been tagged just an hour ago: https://gitlab.alpinelinux.org/alpine/aports/-/commit/86d5892796fb5c817aab0c07a56931bc08d0af61
Release checklist: https://gitlab.alpinelinux.org/alpine/aports/-/issues/16119

DrTimothyAldenDavis:
Is this superseded by https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/810 ?

DrTimothyAldenDavis:
See #810 for the alternative option taken.

PR #805, opened 2024-05-20
Allow turning off SPEX python interface

rayegun:
There's currently no way to turn this off when building shared libs, and it would be cleanest to allow turning it off at the SuiteSparse level, not just the SPEX level.

### DIFF
--- a/.github/workflows/root-cmakelists.yaml
+++ b/.github/workflows/root-cmakelists.yaml
@@ -389,7 +389,7 @@ jobs:
         if: matrix.cuda == 'with'
         id: cuda-toolkit
         with:
-          cuda: '12.2.0'
+          cuda: '12.4.1'
           #See https://docs.nvidia.com/cuda/cuda-installation-guide-microsoft-windows/index.html#install-the-cuda-software
           method: 'local'
           # Do not cache the installer (~3 GiB). It doesn't speed up the
--- a/.github/workflows/root-cmakelists.yaml
+++ b/.github/workflows/root-cmakelists.yaml
@@ -384,12 +384,12 @@ jobs:
 
           msystem: UCRT64
 
-      - uses: Jimver/cuda-toolkit@v0.2.15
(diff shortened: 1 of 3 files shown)


mmuetzel:
Looks good to me.

Should the new CMake parameter be named `SUITESPARSE_USE_PYTHON` (or `SUITESPARSE_ENABLE_PYTHON`) to better fit the naming convention that is being used for recent changes?

rayegun:
Good point, updated the naming

mmuetzel:
It would probably be nice to document the new variable. Maybe, around here?
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/00863cdebe8d2640150ab5c2f087042e6f2858b5/SuiteSparse_config/Config/README.md.in#L982

DrTimothyAldenDavis:
yes, adding something to the top-level README would be a good idea.  Also in the SPEX/README.

I have some other changes to SPEX that are pending, in the paru_dev2 branch.  I'm making some major changes to ParU in that branch, and while I was working on that I found some issues with SPEX when building the MATLAB interface.  I'll rebase this to the paru_dev2 branch and merge it there.

Then I, or @rayegun, could add the README updates to paru_dev2.

Alternatively, I could do a PR of paru_dev2 into dev2, but I'm hesitant to do that just now without resolving some of the other PRs to dev2 just now.

DrTimothyAldenDavis:
Correction:  I'll go ahead and merge this into dev2 and then keep paru_dev2 updated with PRs from dev2 to paru_dev2.  Less churn that way.

Also: I will rename this as `SUITESPARSE_USE_PYTHON` since the `_USE_` style is used for all other options like this one.

DrTimothyAldenDavis:
Updated with this commit:

https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/812/commits/8d59151f13eff38e585141e27b1434e455c1e5ec

PR #807, opened 2024-05-23
CI (macos): Query installation prefix of Homebrew 

mmuetzel:
The `macos-latest` label was changed to point to `macos-14` runners somewhat recently:
https://github.com/github/roadmap/issues/926
Those runners are on Apple Silicon.
For some reason, Homebrew decided to use a different installation prefix on Intel and on Apple Silicon:
https://docs.brew.sh/FAQ#why-should-i-install-homebrew-in-the-default-location

Query the installation prefix of Homebrew and use it where necessary.

Also, add a schedule to trigger the workflow weekly. This can be useful not only to detect possibly breaking changes in SuiteSparse itself. But it can also be useful to detect upstream changes (e.g., in some Homebrew package or GitHub runner configuration) that require changes of the build rules.

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -51,6 +51,8 @@ jobs:
           brew install --overwrite python@3.10 python@3.11 python@3.12
           brew reinstall gcc
           brew install autoconf automake ccache cmake gmp lapack libomp mpfr openblas
+          HOMEBREW_PREFIX=$(brew --prefix)
+          echo "HOMEBREW_PREFIX=$(brew --prefix)" >> $GITHUB_ENV
 
       - name: prepare ccache
         # create key with human readable timestamp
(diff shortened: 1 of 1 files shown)


mmuetzel:
The build issue on the Arch Linux runners is most likely unrelated.
Arch Linux 3.20 has just been tagged a couple of hours ago. That might be a hick-up in the upgrade process that the Arch Linux maintainers will hopefully address soon.

mmuetzel:
I triggered the macos runner on my fork of the repository, and the OpenMP implementation from Homebrew is detected again on Apple Silicon:
https://github.com/mmuetzel/SuiteSparse/actions/runs/9203934509/job/25316416990

mmuetzel:
I opened an issue at the tracker of Alpine Linux for the package installation error on Alpine Linux 3.20:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/16136

PR #809, opened 2024-05-23
Example: Don't include OpenMP header if building without it.

mmuetzel:
Include `omp.h` only if `_OPENMP` is defined.

This fixes an issue when trying to build the example on a system on which no OpenMP implementation can be detected by CMake.

Afaict, the example doesn't use any OpenMP features directly. So, it shouldn't be necessary at all to include the OpenMP header. If a library requires that the OpenMP header is included, that library should include it in its own header.

Should including `omp.h` in the header of the example be removed entirely?

### DIFF
--- a/Example/Include/my_internal.h
+++ b/Example/Include/my_internal.h
@@ -2,7 +2,7 @@
 // SuiteSparse/Example/Include/my_internal.h
 //------------------------------------------------------------------------------
 
-// Copyright (c) 2022-2023, Timothy A. Davis, All Rights Reserved.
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Let's go with this change.  

> Should including `omp.h` in the header of the example be removed entirely?

Perhaps, but for now, let's keep this as modified with the `ifdef`.  That way there's a little less churn in the repo.  I'd rather not break something else just now, since there are a lot of other PRs that are breaking things at the moment (on the mac), or changes in the github runners (alpine for instance).

PR #810, opened 2024-05-23
CI (alpine): Add an emulated RISC-V 64-bit runner

mmuetzel:
This is a simpler alternative to #801 now that Alpine Linux 3.20 is released that includes packages for RISC-V 64.

Like @suvorovrain already wrote in that PR, valgrind isn't packaged for that architecture. (Upstream valgrind doesn't support that architecture.)
I'm not sure why it is installed in the first place on those runners. Could it be removed entirely?

### DIFF
--- a/.github/workflows/build-arch-emu.yaml
+++ b/.github/workflows/build-arch-emu.yaml
@@ -38,19 +38,19 @@ jobs:
         arch: [x86, aarch64, armv7, ppc64le, s390x, riscv64]
         include:
           - arch: x86
-            ccache-max: 80M
+            ccache-max: 64M
             extra-build-libs: ":GraphBLAS:LAGraph"
             extra-check-libs: ":GraphBLAS:LAGraph"
           - arch: aarch64
-            ccache-max: 42M
+            ccache-max: 28M
           - arch: armv7
-            ccache-max: 42M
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
valgrind could be removed, or better yet it could be commented out.  I had it there in case I need to debug something through the github CI, where an error was occuring on some platform I didn't have access to except through the github runners.   Some of the demos and examples can be run with valgrind, by editting the Makefiles.

mmuetzel:
Thank you for clarifying.
I added a commit that comments out the line that would install valgrind. (I hope I didn't mess up the YAML syntax by that change.)

DrTimothyAldenDavis:
Looks good.  Shall I merge it in now or would it best to wait for all the CI's to finish?

mmuetzel:
> Looks good. Shall I merge it in now or would it best to wait for all the CI's to finish?

I believe that file is still valid YAML with the last change. But it would probably be best to wait at least until (and if) the CI starts with that change.

mmuetzel:
It looks like the new runner built and checked without issues. 🎉

(Only slightly related: Still no feedback on the bug report regarding the lapack packages on the aarch64 and ppc64le Alpine Linux runners.)

PR #817, opened 2024-05-24
SPEX: Account for SUITESPARSE_USE_STRICT when using SPEX_USE_PYTHON

mmuetzel:
Handle SPEX_USE_PYTHON similar to SPEX_USE_OPENMP when SUITESPARSE_USE_STRICT is ON.

Another addendum to #805.

### DIFF
--- a/SPEX/CMakeLists.txt
+++ b/SPEX/CMakeLists.txt
@@ -415,7 +415,13 @@ endif ( )
 # SPEX_USE_PYTHON (for just SPEX) or SUITESPARSE_USE_PYTHON (for all of
 # SuiteSparse).  Currently, only SPEX has a Python interface in SuiteSparse.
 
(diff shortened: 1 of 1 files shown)


mmuetzel:
The latest commit uses a default value of `OFF` for `SPEX_USE_PYTHON` if `BUILD_SHARED_LIBS` evaluates to false.
I'm wondering if it would be better to make the default value of `SUITESPARSE_USE_PYTHON` dependent on `BUILD_SHARED_LIBS` instead...

What do you think?

mmuetzel:
Re-opening to trigger the CI again. The issue with the package signatures from Alpine Linux should be fixed now:
https://gitlab.alpinelinux.org/alpine/aports/-/issues/16136

DrTimothyAldenDavis:
Looks great -- thanks!

PR #823, opened 2024-06-03
UMFPACK: Display timer function used by SuiteSparse_config in report

### LINKED ISSUE #822: CPU timer: no timer used.
+ cmake command
`cmake -G"Ninja Multi-Config" -DCMAKE_CXX_COMPILER=cl -DCMAKE_C_COMPILER=cl -DCMAKE_Fortran_COMPILER=ifort -DSUITESPARSE_USE_OPENMP=ON -DSUITESPARSE_ENABLE_PROJECTS="umfpack" -DCMAKE_INSTALL_PREFIX=F:\umfpackWork\suitesparse7.7 --fresh ..`
+ compile command
cmake --build . --config Release

+ test code
```c++
#include <stdio.h>
#include "umfpack.h"
int n = 5;
int Ap[] = {0, 2, 5, 9, 10, 12};
int Ai[] = {0, 1, 0, 2, 4, 1, 2, 3, 4, 2, 1, 4};
double Ax[] = {2., 3., 3., -1., 4., 4., -3., 1., 2., 2., 6., 1.};
double b[] = {8., 45., -3., 3., 19.};
double x[5];
int main(void)
{
    double *null = (double *)NULL;
    int i;
    void *Symbolic, *Numeric;
    double infos[UMFPACK_INFO];
    (void)umfpack_di_symbolic(n, n, Ap, Ai, Ax, &Symbolic, null, infos);
    (void)umfpack_di_numeric(Ap, Ai, Ax, Symbolic, &Numeric, null, infos);
    umfpack_di_free_symbolic(&Symbolic);
    (void)umfpack_di_solve(UMFPACK_A, Ap, Ai, Ax, x, b, Numeric, null, infos);
    umfpack_di_free_numeric(&Numeric);
    for (i = 0; i < n; i++)
        printf("x [%d] = %g\n", i, x[i]);
    double control[UMFPACK_CONTROL];
    umfpack_di_defaults(control);
    control[UMFPACK_PRL] = 7;
    umfpack_di_report_info(control,infos);
}
```
output
```shell
x [0] = 1
x [1] = 2
x [2] = 3
x [3] = 4
x [4] = 5
UMFPACK V6.3.3 (Mar 22, 2024), Info:
    matrix entry defined as:          double
    Int (generic integer) defined as: int32_t
    BLAS library used: Intel MKL 64lp BLAS (32-bit integers).  size of BLAS integer: 4
    MATLAB: no.
    CPU timer: no timer used.
    number of rows in matrix A:       5
    number of columns in matrix A:    5
    entries in matrix A:              12
...
```

But if I set `-DCMAKE_C_FLAGS=-openmp`, timer will be set to omp_get_wtime()
```shell
CPU timer: omp_get_wtime ( )
```
Why `-DSUITESPARSE_USE_OPENMP=ON` not work?

mmuetzel:
UMFPACK is built without OpenMP. (It doesn't use it directly.) Hence, `_OPENMP` is not defined when the UMFPACK libraries are built.

Get the information about the used timer function directly from SuiteSparse_config.h instead.

Fixes #822.

I used the following reference for the "stringification" of the preprocessor macro:
https://gcc.gnu.org/onlinedocs/cpp/Stringizing.html


mmuetzel:
Oops. What I wrote is probably not entirely correct.

I was assuming that all libraries would be using `SuiteSparse_tic`, `SuiteSparse_toc` or `SuiteSparse_time` to measure time. However, it looks like some libraries (or demos) are using `SUITESPARSE_TIME` directly. For that to work, they need to build with OpenMP (if `clock_gettime` is not available).

Additionally, the proposed change doesn't do what I thought it would because `SUITESPARSE_TIME` is set to a value that depends on the library in which `SuiteSparse_config.h` is included (and not on how libsuitesparse_config itself was built).

Should `SUITESPARSE_TIME` really be used in libraries that depend on libsuitesparse_config? Should these libraries (and demos) better use `SuiteSparse_time` instead?

mmuetzel:
I added a second commit that should be fixing the issues described in the previous comment.

mmuetzel:
It looks like the changes from this PR triggered a test error on the runner using `clang-cl`. 
It is likely that there is an issue with the OpenMP configuration on that runner indeed. It is linking libraries that are using the Intel OpenMP implementation (the Intel MKL BLAS/LAPACK libraries) and libraries that are using the LLVM OpenMP implementation (the SuiteSparse libraries).
That might also be an issue on the runners that are using MSVC `cl`. But it looks like they do not have a similar runtime check in the MSVC implementation of OpenMP.

I'll try to look into that soon. (But that will likely result in a different PR.)

mmuetzel:
Converted to draft for now to wait for #824 and rebase.

mmuetzel:
This seems to be working now.

It looks like GitHub is in the process of updating MSVC from 19.39 to 19.40 on their runners. And the CUDA version that we are installing isn't compatible with that newer version of MSVC.
But that is unrelated to the changes proposed here.

PR #826, opened 2024-06-05
CI (MSVC): Update build rules for MSVC 19.40

mmuetzel:
GitHub seems to be updating their runner images to include MSVC 19.40 currently.

NVIDIA CUDA seems to think this newer version of MSVC is no longer Microsoft Visual Studio 2022 (but it is).
Work around that issue by adding `-allow-unsupported-compiler` to the CUDA flags.

Additionally, it looks like the PATH variable on these runner images is arranged such that a different (older) `clang-cl` is being picked up by CMake instead of the newer one from MSVC 19.40.
Work around that by moving that version out of the PATH so CMake picks the compiler that is installed with Microsoft Visual Studio.
See also https://github.com/actions/runner-images/issues/10001.


mmuetzel:
Now the MSYS2 mirrors are timing out... What's going on today?

I'll try to force push later to see if it will work then.

mmuetzel:
Looks like the CI is green again with this change.

Issue #835, opened 2024-06-11
Building SuiteSparse for use on iOS platform, fails on coding signing step of JITpackage.


sbelbin:
**Describe the bug**
Building SuiteSparse to target the iOS platform such as Apple's iPad the Cmake build fails at code signing phase for the **JITpackage**.

**To Reproduce**
1. Clone the **ios-cmake** toolchain from here https://github.com/leetal/ios-cmake in a folder above SuiteSparse folder.
2. Download and build **GMP**, **MPFR** and **OpenMP** and build it for the **iOS** platform.
3. Clone the **SuiteSparse** from GitHub.
4. cd SuiteSparse
5. cmake -B build -G Xcode --toolchain ../ios-cmake/ios.toolchain.cmake -DPLATFORM=OS64 -DCMAKE_INSTALL_PREFIX=../install/OS64/suitesparse .
6. cmake --build ./build

**Expected behavior**
Expectation that the **JITpackage** to build and for code signing to occur.

**Screenshots**
/Users/yul/tmp/SuiteSparse/build/GraphBLAS/JITpackage/native/jitpackage.xcodeproj: error: Signing for "grb_jitpackage" requires a development team. Select a development team in the Signing & Capabilities editor. (in target 'grb_jitpackage' from project 'jitpackage')

**Desktop (please complete the following information):**
 - OS: [iOS]
 - compiler [Apple clang version 15.0.0]
 - BLAS and LAPACK library, if applicable
 - Version [v7.7.0]

**Additional context**
Our development environment is a MacBook Pro. Our C/C++ library is build using Visual Studio Code along with CMake, whereas the iPad app is written in Swift in Xcode with a bridging header to integrate our C library to the iPad app.

The **SuiteSparse** dependencies of **GMP**, **MPFR** and **OpenMP** have been built the **iOS** platform and the SuiteSparse code is referring to the appropriate **include** and **lib** folders.

I've tried disabling the building of shared library via the **-DBUILD_SHARED_LIBS=OFF** option on the CMake comand line. Hoping that since we are building to be a static library that it would forego this code signing step.

Additionally, I've tried setting these CMake definitions below hoping what was missing was a certificate to use when signing the code.

set ( CPACK_BUNDLE_APPLE_CERT_APP "Apple Development: FirstName LastName (XX9X9X9XXX)")
set ( CMAKE_XCODE_ATTRIBUTE_DEVELOPMENT_TEAM "XX9X9X9XXX" )
[...]

PR #836, opened 2024-06-11
GraphBLAS: Check if C standard function "system" can be used

mmuetzel:
Support for the CPU JIT compiler requires that a compiler can be executed on the host using the `system` function.

Disable the JIT compiler if the `system` function cannot be used on the target system.

See also the discussion in #835.

### DIFF
--- a/GraphBLAS/cmake_modules/GraphBLAS_JIT_paths.cmake
+++ b/GraphBLAS/cmake_modules/GraphBLAS_JIT_paths.cmake
@@ -43,8 +43,20 @@ if ( SUITESPARSE_HAS_CUDA AND GRAPHBLAS_USE_CUDA )
 #   set ( GRAPHBLAS_COMPACT on )    # FIXME
 endif ( )
 
+include ( CheckSymbolExists )
+check_symbol_exists ( system "stdlib.h" HAVE_C_SYSTEM )
+
 option ( GRAPHBLAS_COMPACT "ON: do not compile FactoryKernels.  OFF (default): compile FactoryKernels" OFF )
-option ( GRAPHBLAS_USE_JIT "ON (default): use the CPU JIT.  OFF: do not use the CPU JIT" ON )
+option ( GRAPHBLAS_USE_JIT "ON (default): use the CPU JIT.  OFF: do not use the CPU JIT" ${HAVE_C_SYSTEM} )
+
+if ( GRAPHBLAS_USE_JIT AND NOT HAVE_C_SYSTEM )
+    if ( SUITESPARSE_USE_STRICT )
+        message ( FATAL_ERROR "GRAPHBLAS CPU JIT cannot be built without support for the C standard function 'system'." )
+    else ( )
+        message ( STATUS "C standard function 'system' cannot be used. GRAPHBLAS CPU JIT will be disabled." )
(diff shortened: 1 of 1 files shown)


mmuetzel:
Wait a second. Not quite right yet.

mmuetzel:
There was an error in the syntax for overriding a cache variable.
Should be fine now.

GraphBLAS/cmake_modules/GraphBLAS_JIT_paths.cmake
```diff
@@ -43,8 +43,20 @@ if ( SUITESPARSE_HAS_CUDA AND GRAPHBLAS_USE_CUDA )
...
 endif ( )
 
+include ( CheckSymbolExists )
+check_symbol_exists ( system "stdlib.h" HAVE_C_SYSTEM )
+
 option ( GRAPHBLAS_COMPACT "ON: do not compile FactoryKernels.  OFF (default): compile FactoryKernels" OFF )
-option ( GRAPHBLAS_USE_JIT "ON (default): use the CPU JIT.  OFF: do not use the CPU JIT" ON )
+option ( GRAPHBLAS_USE_JIT "ON (default): use the CPU JIT.  OFF: do not use the CPU JIT" ${HAVE_C_SYSTEM} )
+
+if ( GRAPHBLAS_USE_JIT AND NOT HAVE_C_SYSTEM )
```
sbelbin, line 52: It seems odd that at line 50 the **GRAPH_USE_JIT** option is set according to if the above symbol check for system passes then to verify **HAVE_C_SYSTEM**.

To me it seems that **if ( GRAPHBLAS_USE_JIT )** does the job. Otherwise keep this line 52 as is but remove the line 50.

I'll state that the check_symbol_exists from the stdlib.h is a nice technique. I've forgotten about this approach in CMake. I like it since it adapts to potentially other platforms & scenarios. Good job.

DrTimothyAldenDavis: The GRAPHBLAS_USE_JIT is a user-selectable option. It tells use "the end user would like to use the JIT in GraphBLAS".  The default setting is ON if the system() call is available (the normal case).

If the setting is ON but the JIT fails, then we either (a) disable the JIT, or (b) report an error.  The latter occurs if SUITESPARSE_USE_STRICT is ON.  That's what line 52 is checking (and line 53).

Normally for these kind of options we have `GRAPHBLAS_USE_something` (a request from the end user) and `GRAPHBLAS_HAS_something` (if the option has been requested and successfully fulfilled).  So something can be requested (with `_USE_` being ON) but we can't do it (`_HAS_` is OFF).

DrTimothyAldenDavis: The GRAPHBLAS_USE_JIT is just a single variable but that looks fine to me.

mmuetzel: The last argument of `option` in line 50 is the *default* value for `GRAPHBLAS_USE_JIT`. The user might have set a different value, e.g., by adding `-DGRAPHBLAS_USE_JIT=ON` in the configure step. That would override the default value.

Line 52 is there to check if the build system should react on such a "known bad" configuration.

PR #876, opened 2024-10-18
Use same objects for linking to shared and static libraries.

mmuetzel:
Avoid building all objects twice (for the shared and the static library), effectively reducing the build time approximately by a factor of two.

Something similar could probably be done for most of the SuiteSparse libraries (including GraphBLAS). So, build time could probably be reduced by a lot (when building shared and static libraries).

### DIFF
--- a/AMD/CMakeLists.txt
+++ b/AMD/CMakeLists.txt
@@ -66,13 +66,7 @@ configure_file ( "Config/amd_version.tex.in" "${PROJECT_SOURCE_DIR}/Doc/amd_vers
     NEWLINE_STYLE LF )
 
 #-------------------------------------------------------------------------------
-# include directories
-#-------------------------------------------------------------------------------
-
-include_directories ( Source Include )
-
-#-------------------------------------------------------------------------------
-# dynamic amd library properties
+# object library
 #-------------------------------------------------------------------------------
 
 if ( SUITESPARSE_HAS_FORTRAN )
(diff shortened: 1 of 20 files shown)


DrTimothyAldenDavis:
Looks great!  That came out very clean.

My organization requires a minor revision to my Contributor Agreement, for all new contributions.  Can you sign the new one I just posted?  It's here in the dev branch:

mmuetzel:
I applied the same idea to some other SuiteSparse libraries and pushed here.

It looks like that doesn't work in general though. 😓
Maybe, position-independent code in static vs. shared libraries? On some platforms only? 🤔

mmuetzel:
Explicitly compiling with `-fPIC` seems to fix it. (At least, it is able to link `libsuitesparseconfig.so` on Linux with that change.)

Would that be an ok fix? IIUC, static libraries are built without `-fPIC` by default...

Edit: I wonder why that wasn't necessary for libamd.so...

DrTimothyAldenDavis:
> Explicitly compiling with `-fPIC` seems to fix it. (At least, it is able to link `libsuitesparseconfig.so` on Linux with that change.)
> 
> Would that be an ok fix? IIUC, static libraries are built without `-fPIC` by default...
> 
> Edit: I wonder why that wasn't necessary for libamd.so...

Oh ... that could be a show-stopper.  I think I shouldn't create static libraries with -fPIC.  That could really be backward breaking change to the many users of the SuiteSparse static libraries.  There's no telling what could break.  It seems very dangerous to go down this path.

Adding -fPIC to static libraries can make them slower, for example.  I think there's a reason that some people want to use static libraries and avoid -fPIC.  The reasons might be more than just performance.

PR #880, opened 2024-10-27
Reduce compiler warnings (-Wall -Wpedantic)

gruenich:


DrTimothyAldenDavis:
Looks good.  I haven't had time to try these out but will do so when my schedule allows.

My organization requires a minor revision to my Contributor Agreement, for all new contributions.  Can you sign the new one I just posted?  It's here in the dev branch:

DrTimothyAldenDavis:
Also, I need to revise the PR to go to the dev2 branch, not dev.  The two branches are in sync so that shouldn't be a problem.

Mongoose/Tests/Mongoose_Test_EdgeSeparator.cpp
```diff
@@ -65,8 +65,8 @@ int runEdgeSeparatorTest(const std::string &inputFile, const double targetSplit)
         int count = 0;
         for (int i = 0; i < result->n; i++)
         {
-            bool equals_0 = (result->partition[i] == 0);
-            bool equals_1 = (result->partition[i] == 1);
+            [[maybe_unused]] bool equals_0 = (result->partition[i] == 0);
```
imciner2: The `[[maybe_unused]]` attribute was only added to the C++ standard in C++17, but the Mongoose CMake is configured to request C++11
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/fd8a7d1db63557a8262793f0c106f6e3dcf86581/Mongoose/CMakeLists.txt#L192
So I think this could lead to problems with some compilers if they are stricter with standard conformance.

DrTimothyAldenDavis: Good point -- thanks for the catch!

Mongoose only assumes c++11 so we cannot use that here.

gruenich: You are right. This is sad, I was happy to get this fixed for C++ so easily. Let us wait for C++17 and C23!

gruenich:
I think this is ready to be merged.

DrTimothyAldenDavis:
I haven't had a chance to get to it yet.  Each of these changes looks fine but I need to evaluate each one carefully.  Sometimes they point to a deeper issue in the code, which I have to think about.

I'm in the process of doing the same for GraphBLAS.  I just encountered a bug where I failed to return a value from a non-void function.  It worked, surprisingly, with -g but failed with -O3.  Took a while to track down.  So I've turned on -Wreturn-type as an error, so the compiler will refuse to compile any code that does that.

I may need to do things like that in the rest of SuiteSparse, and some of these might be just those cases.

Unused variables are different.  Sometimes I want to keep them, for code readability, as a comment.  In other words, it might be a line of code I want to see but don't actually need to compile.  But in that case I should comment them out, not delete them.

DrTimothyAldenDavis:
And sometimes unused variables are actually used if debugging is enabled with either #undef NDEBUG or my own debug structure.  I have to double check to make sure which ones are used for debugging.  In that case, the variables need to be placed inside and #ifdef so they are available when debugging is enabled.

So I need to check each one.

This is a useful process but it takes some time to evaluate each of these cases.

DrTimothyAldenDavis:
OK -- looks great.  I'm in the process of syncing SuiteSparse with some of the other repos it holds a copy of.  SPEX and LAGraph have their own repos elsewhere.  I'll copy these changes over to the appropriate branches there.

PR #894, opened 2025-01-29
CI: Add Ubuntu on ARM runner to CI

mmuetzel:
GitHub started hosting runners with Ubuntu on ARM64 processors for open source projects for free:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

Add one configuration that is using these runners to the build matrix.

According to their blog post, the arm64 runners are more efficient and potentially faster than the x86_64 runners. (But it is still in a preview phase and maybe it will take some time for them to better scale and balance the load.) If that turns out to be true, it should be easy to switch more configurations in that workflow (or the one using the root CMake file) to the arm64 runners (and maybe keep only one or two running on x86_64).

Since there is now the possibility to run tests for this target natively, remove that target from the build matrix of the workflow that emulates it with `qemu`.

There is an open PR for the Alpine action that would allow running the armv7 tests natively on aarch64. But it hasn't been merged yet:
https://github.com/jirutka/setup-alpine/pull/22
So, keep emulating armv7 on x86_64 for the time being.

### DIFF
--- a/.github/workflows/build-arch-emu.yaml
+++ b/.github/workflows/build-arch-emu.yaml
@@ -35,14 +35,12 @@ jobs:
       matrix:
         # For available CPU architectures, see:
         # https://github.com/marketplace/actions/setup-alpine-linux-environment
-        arch: [x86, aarch64, armv7, ppc64le, s390x, riscv64]
+        arch: [x86, armv7, ppc64le, s390x, riscv64]
         include:
           - arch: x86
             ccache-max: 64M
             extra-build-libs: ":GraphBLAS:LAGraph"
             extra-check-libs: ":GraphBLAS:LAGraph"
-          - arch: aarch64
-            ccache-max: 28M
           - arch: armv7
             ccache-max: 25M
           - arch: ppc64le
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -49,19 +49,14 @@ jobs:
             openmp: with
             link: both
           - os: ubuntu-latest
-            compiler: clang
-            cuda: with
-            openmp: with
-            link: both
-          - os: ubuntu-latest
-            compiler: clang
-            cuda: without
-            openmp: with
-            link: both
-          - os: ubuntu-latest
             compiler: gcc
             cuda: without
             openmp: without
+            link: both
+          - os: ubuntu-latest
+            compiler: clang
+            cuda: with
+            openmp: with
             link: both
           - os: ubuntu-latest
             compiler: gcc
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -23,47 +23,59 @@ jobs:
   ubuntu:
     # For available GitHub-hosted runners, see:
     # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
 
-    name: ubuntu (${{ matrix.compiler }} ${{ matrix.cuda }} CUDA ${{ matrix.openmp }} OpenMP, ${{ matrix.link }})
+    name: ${{ matrix.os }} (${{ matrix.compiler }} ${{ matrix.cuda }} CUDA ${{ matrix.openmp }} OpenMP, ${{ matrix.link }})
 
     strategy:
       # Allow other runners in the matrix to continue if some fail
       fail-fast: false
 
       matrix:
+        os: [ubuntu-latest]
         compiler: [gcc]
         cuda: [with]
         openmp: [with]
         link: [both]
         include:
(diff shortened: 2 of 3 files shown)


mmuetzel:
Hu? Why are the runners that are using Clang on Ubuntu failing? I didn't touch anything in the rules about them...

Looking back on the actions that ran in the last couple of weeks: That isn't new with the changes from here. These runners started failing without any change in this repository.
Might be related to the fact that the `ubuntu-latest` label was moved from `ubuntu-22.04` to `ubuntu-24.04` by GitHub recently. That came with a newer version of Clang...

A few tests for LAGraphX are failing with a segfault now.
@DrTimothyAldenDavis: Do you see a pattern for the failing tests?

```
  The following tests FAILED:
  	 39 - LAGraphX_BF (SEGFAULT)
  	 40 - LAGraphX_FastGraphletTransform (SEGFAULT)
  	 50 - LAGraphX_SquareClustering (SEGFAULT)
  	 55 - LAGraphX_msf (Failed)
  Errors while running CTest
  make: *** [Makefile:74: test] Error 8
```

mmuetzel:
Sure enough, those workflows succeed if they are reverted to the Ubuntu 22.04 runners.
That could be a work-around for the time being. But it might also be worth finding out *why* it is failing on Ubuntu 24.04. (Fwiw, GitHub - at least until now - provides runners for the last two LTS versions of Ubuntu. So, the Ubuntu 22.04 runners will probably be available until some time in 2026.)

What I've missed before: The runners that build for Windows MSVC using Clang as the compiler are also failing the tests for LAGraph.
It might be that the compatibility with newer versions of LLVM Clang is broken for some reason (independent on the OS -- Linux or Windows).

DrTimothyAldenDavis:
Thanks.  yes, I've noticed the Windows clang failures.  I was considering just deleting those runners.

mmuetzel:
I tried to enable CUDA on the Ubuntu on ARM runner. But CMake couldn't detect a working CUDA compiler (albeit installing the same packages that work for Ubuntu on x86_64).
```
  -- Looking for a CUDA compiler
  -- Looking for a CUDA compiler - NOTFOUND
  -- CUDA:             not found
  CMake Error at cmake_modules/SuiteSparsePolicy.cmake:367 (message):
    CUDA required for SuiteSparse but not found
    CMakeLists.txt:46 (include)
  
  
  -- CUDA:             not enabled
  -- Configuring incomplete, errors occurred!
  Error: Process completed with exit code 1.
```

I don't know how to install a working CUDA toolchain on ARM or how to tell CMake to use it. I also can't test that locally (no ARM hardware here). So, I opted for disabling CUDA on the ARM runner.

In the last run for the root CMakeLists.txt file on Ubuntu on ARM, checking out the repository failed apparently for some reason.
@DrTimothyAldenDavis: When you come around to it, could you please restart that runner?

Other than that, this should be ready for review now.

mmuetzel:
Failing to checkout the repository intermittently on ARM runners seems to be a known issue:
https://github.com/actions/partner-runner-images/issues/47

Changing to draft until the image providers have figured out a solution for that.

mmuetzel:
> Changing to draft until the image providers have figured out a solution for that.

Apparently, that was caused by a hardware error(?) that they worked around by switching out their runners:
https://github.com/orgs/community/discussions/148648#discussioncomment-12205917

Marking as ready for review again.

DrTimothyAldenDavis:
Looks great.

I hope to release a new version of SuiteSparse soon (v7.9.0).  It will include these changes, and the other pending PRs to SuiteSparse once I review them and integrate them.  I have a minor bug fix to post in GraphBLAS as well (v9.4.5, currenly in the v9.4.branch in the GraphBLAS repo).

Then probably in about 2 weeks I'll post a GraphBLAS v10.0.0, and add it to SuiteSparse v7.10.0.

DrTimothyAldenDavis:
The segfaults are because of a clang compiler bug, which affects GraphBLAS/Source/mask/GB_masker.c.

The fix is here, just posted on dev2 a few hours ago:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blame/ece49b5f114771b2eeafefd3edb3f86714ec7e61/GraphBLAS/Source/mask/GB_masker.c#L52-L64

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/ece49b5f114771b2eeafefd3edb3f86714ec7e61/GraphBLAS/Source/mask/GB_masker.c#L52-L63

That should allow clang to properly compile this file.  See also this note in the GraphBLAS/Doc/Changelog:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/ece49b5f114771b2eeafefd3edb3f86714ec7e61/GraphBLAS/Doc/ChangeLog#L6-L20

Issue #895, opened 2025-02-01
cmake built error in static win msvc intel due linking static.lib : wrongname and cant linking mkl when create 


realbabilu:
**Describe the bug**
Microsoft (R) Program Maintenance Utility Version 14.42.34436.0
Copyright (C) Microsoft Corporation.  All rights reserved.

[  0%] Built target SuiteSparseConfig
[  1%] Built target SuiteSparseConfig_static
NMAKE : fatal error U1073: don't know how to make 'SuiteSparse_config\suitesparseconfig.lib'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.42.34433\bin\HostX64\x64\nmake.exe" -s -f AMD\CMakeFiles\AMD.dir\build.make /nologo -SL                 AMD\CMakeFiles\AMD.dir\build' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.42.34433\bin\HostX64\x64\nmake.exe" -s -f CMakeFiles\Makefile2 /nologo -S                  all' : return code '0x2'
Stop.
;
it should be 'SuiteSparse_config\suitesparseconfig.lib' SuiteSparse_config\suitesparseconfig._static.lib'
**To Reproduce**
Steps to reproduce the behavior.
cmake -G "NMake Makefiles"  -DBLA_VENDOR=Intel10_64lp -DCMAKE_BUILD_TYPE=RELEASE -DSUITESPARSE_ENABLE_PROJECTS="suitesparse_config;amd;umfpack" -DCHOLMOD_PARTITION=OFF -DBUILD_STATIC_LIBS=ON -DSUITESPARSE_USE_FORTRAN=on ..   -DCMAKE_CXX_COMPILER=icx.exe

**Expected behavior**
in build.make in AMD.dir should be

AMD\amd.dll: SuiteSparse_config\suitesparseconfig_static.lib

[...]

PR #896, opened 2025-02-03
CMake: Adjust condition for changed name for static libraries.

mmuetzel:
MSVC-like compilers use the same default suffix for static libraries and import libraries.
Instead of checking whether a compiler is MSVC or simulates it, directly check whether these default suffices clash.

That potentially fixes issue #895. (And even in case it does not, the updated condition is much more to the point.)


mmuetzel:
Turns out that #895 was probably caused by an issue with NMake (i.e., most likely upstream).

Closing this PR because it didn't change anything.

PR #901, opened 2025-02-15
CI: Overhaul CI rules for MSVC and disable GraphBLAS JIT compiler with clang

mmuetzel:
Some of the CI jobs that are targeting windows-msvc using Clang as the compiler are currently failing.

To be able to have more granularity when iterating on the CI rules, split the rules for windows-msvc to a separate file.
Also, simplify the build matrix by moving more of the complexity to the build steps. (Similar to what was recently done for the CI rules for Ubuntu.)

It turns out that disabling the JIT compiler of GraphBLAS avoids the test failures for the job that is using `clang` as the compiler driver. I don't know *why* that is though. But people that would prefer using the JIT compiler on Windows could still use Microsoft's own `cl` as the compiler (or build for windows-gnu, i.e., MinGW).

If I understand correctly, using `clang` or `clang-cl` as the compiler driver shouldn't make much difference for the resulting built artifacts. I thought, the main (and only?) difference is that `clang` takes arguments that are more compatible to `gcc` and `clang-cl` takes arguments that are compatible with `cl`.
But still, some of the tests are failing (even without the JIT compiler) when `clang-cl` is used as the compiler driver. (However, it is failing a lot less tests with disabled JIT compiler.)
I tried disabling OpenMP or building in debug mode. But none of that made a difference.

Somewhat unrelated to the remaining changes in this PR:
The compiler was complaining that the function `GB_enumify_masker` doesn't return a value while its prototype indicates a return type of `uint64_t`.
I couldn't find anywhere in GraphBLAS or LAGraph where an output argument for that function was used. So, I opted for changing the function prototype to "return" `void`. I hope that is OK.


DrTimothyAldenDavis:
Thanks!  I'll double-check the GB_enumify_masker.  I may have already resolved that in the draft of v10, which I'm in the thick of right now.  Or, in the draft of v9.4.4 which is close to being released (being a minor bug update version).  I'll incorporate this change into v9.4.4 if I haven't already done so.

mmuetzel:
> I'll incorporate this change into v9.4.4 if I haven't already done so.

Afaict, you didn't include that change in the GraphBLAS 9.4.4 release:
https://github.com/DrTimothyAldenDavis/GraphBLAS/blob/31a1c6bdc00b67fe863712e89283618cfad8ac07/Source/jitifyer/GB_enumify_masker.c#L13

Was that intentional?
Would it help if I opened a PR with that change in your GraphBLAS repository?

DrTimothyAldenDavis:
Oops -- that was an oversight.  I have too many codes under revision at the same time ...

I also found a bug in v9.4.4 that is now fixed in the v9.4.branch of the GraphBLAS stand-alone repo: https://github.com/DrTimothyAldenDavis/GraphBLAS/tree/v9.4.branch 

I can make the change there, or you can do a PR to the v9.4.branch with the GraphBLAS changes.

DrTimothyAldenDavis:
I'll go ahead and merge this into SuiteSparse now.

DrTimothyAldenDavis:
The enumify-masker change looks fine -- I went ahead and added it to v9.4.branch.  I will add this to v9.4.5 soon.

Issue #799, opened 2024-05-09
compiled failed on windows


DragonPara:
**Describe the bug**
Can't compile CHOLMOD

**To Reproduce**
mkdir build
cd build
`cmake ..  -DCMAKE_CXX_COMPILER=icc -DCMAKE_C_COMPILER=icc -DCMAKE_Fortran_COMPILER=ifort -DCMAKE_INSTALL_PREFIX="E:/umfpackCompilerWork/SuiteSparse7.7" -DSUITESPARSE_ENABLE_PROJECTS="umfpack"`
the result of cmake stored in `config.log`
[config.log](https://github.com/DrTimothyAldenDavis/SuiteSparse/files/15258994/config.log)
`msbuild SuiteSparse.sln /p:Configuration=Release > compile.log`
the result of build stored in compile.log
[compile.log](https://github.com/DrTimothyAldenDavis/SuiteSparse/files/15258993/compile.log)

**Expected behavior**
normal compilation

**Screenshots**
![image](https://github.com/DrTimothyAldenDavis/SuiteSparse/assets/44524468/cd13d583-5736-4ae0-b5d8-d23a9cd8218f)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - compiler MSVC
 - intel mkl
 - Version 7.7.0

**Additional context**
I found `-DCMAKE_CXX_COMPILER=icc -DCMAKE_C_COMPILER=icc` also not work.

AuroraLHL:
I meet this error on Mac,how do you fixed it ?

mmuetzel:
I tried to reproduce the error with the following steps on Windows 11 23H2 (but it built without error for me):
* Download and install MSVC 2022 (version 17.9.6)
* Download and install Intel oneAPI 2024.1
* Open the "Intel oneAPI command prompt for Intel 64 for Visual Studio 2022" shell with the shortcut in the Windows Start menu.
* Run the following commands (`icc` or `ifort` didn't install for me with the steps described above):
```
cmake -DCMAKE_CXX_COMPILER=icx-cl -DCMAKE_C_COMPILER=icx-cl -DSUITESPARSE_ENABLE_PROJECTS="umfpack" --fresh ..
cmake --build . --config Release    
ctest . -C Release
```

Could you please describe the exact steps to reproduce the error that you are seeing?

DragonPara:
> I meet this error on Mac,how do you fixed it ?

[...]

PR #931, opened 2025-03-25
CHOLMOD: Use CUDA as linker language if applicable

mmuetzel:
Use CUDA as the linker language if applicable while still avoiding Fortran as the linker language (see #799).

Addresses #929.

I seem to recall that there were CI runners that tested with CUDA using MSVC on Windows. It looks like there are none any more.
Also add one runner that tests building with CUDA on Windows (again?).

### DIFF
--- a/.github/workflows/root-cmakelists-msvc.yaml
+++ b/.github/workflows/root-cmakelists-msvc.yaml
@@ -23,7 +23,7 @@ jobs:
     # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners
     runs-on: windows-latest
 
-    name: msvc (${{ matrix.cc }} ${{ matrix.openmp }} OpenMP, ${{ matrix.link }})
+    name: msvc (${{ matrix.cc }} ${{ matrix.openmp }} OpenMP ${{ matrix.cuda }} CUDA, ${{ matrix.link }})
 
     defaults:
       run:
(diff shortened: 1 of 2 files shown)


DrTimothyAldenDavis:
The cuda CI on MSVC was failing to load the cuda package-- a github or other bug, not mine-- so I had to disable it

mmuetzel:
> The cuda CI on MSVC was failing to load the cuda package-- a github or other bug, not mine-- so I had to disable it

I see.
I seem to recall that downloading the installer from the NVIDIA servers was slow sometimes for the GitHub runners. It still takes its time. But given its size (several Gigabyte), that is probably to be expected.

I'll try to think about what could be done to not fail the CI in case the NVIDIA servers are down (or slow) again.

In the meantime, is it ok to add a runner with CUDA on Windows again?

DrTimothyAldenDavis:
Yes, if it can work let's add it back in

mmuetzel:
I force pushed to trigger the CI again. 
If the installer is downloading also the second time, I'd say that it's worth enabling it for one runner again.

mmuetzel:
The last two commits set `CUDA_RESOLVE_DEVICE_SYMBOLS` to `ON` for the static CHOLMOD library to work around an issue where some CUDA symbols aren't resolved correctly when linking that static library in dependent binaries with MSVC.
That has some side effects as explained in the commit message. But those side effects are probably not important for the CHOLMOD library.
I wonder if that flag should only be set when building with MSVC (or MSVC-compatible compilers).

The last of these two commits adds a runner configuration to the CI matrix for MSVC that should be affected by the issue from #929.

mmuetzel:
I had a look at the functions that are compiled to GPU code (in `cholmod_gpu_kernels.cu`). None of them are declared in the installed header.
I gather that means that they aren't expected to be callable directly from code that is linking to the built CHOLMOD library. (But only from within the CHOLMOD library itself.)

In that case, resolving the CUDA device symbols when linking the static CHOLMOD library shouldn't be an issue. So, the changes in this PR should be OK.

PR #935, opened 2025-04-08
Support building SuiteSparse_config without BLAS.

### LINKED ISSUE #934: Building SuiteSparse crashes when no BLAS library is installed even if the modules chosen by the user do not need it
**Describe the bug**
No matter which SuiteSparse modules the users wishe to build, if they don't have a BLAS library installed on their machine, compilation will fail because the suitesparse_config module always looks for this dependency on the system.

Let us be a user who wants to build SuiteSparse with only a few modules. For example, they need only suitesparse_config, AMD, COLAMD, BTF and KLU. Also, they do not need KLU to use CHOLMOD. Therefore, 

**To Reproduce**

1. Clone the repo
2. Follow the build instructions with these options: `cmake -DSUITESPARSE_ENABLE_PROJECTS="suitesparse_config;amd;colamd;btf;klu" -DKLU_USE_CHOLMOD:BOOL=OFF ..`

**Expected behavior**
I expect the build not to crash due to BLAS being not found since those modules don't need it.

**Screenshots**

 - OS: **Ubuntu 24.04**
 - compiler: **gcc 13**
 - BLAS and LAPACK library, if applicable: **NONE**

**Additional context**
In order to make my build work, I currently have to manually patch `SuiteSparse_config/CMakeLists.txt` and `SuiteSparse_config/SuiteSparse_config.c` to avoid checking for the BLAS dependency.

See https://github.com/powsybl/powsybl-metrix/pull/207 (still a WIP as of 2025/04/08).

mmuetzel:
BLAS is a dependency of SuiteSparse_config for some helper functions that can be used in other SuiteSparse libraries. If no SuiteSparse library that depends on BLAS is build, these helper functions are not used, and they are allowed to return anything in general without potential adverse side effects.

Add a new configuration flag `SUITESPARSE_REQUIRE_BLAS` that can be used to manually remove the requirement for a BLAS implementation when building SuiteSparse_config.

If a user sets `SUITESPARSE_REQUIRE_BLAS` to OFF and attempts to build a library that requires a BLAS library using the root `CMakeLists.txt` file, an error message is shown during configuration.

Fixes #934.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,7 @@ endforeach ( )
 
 # CHOLMOD options affecting dependencies
 option ( CHOLMOD_CAMD "ON (default): use CAMD/CCOLAMD.  OFF: do not use CAMD/CCOLAMD" ON )
+option ( CHOLMOD_SUPERNODAL "ON (default): use Supernodal Module.  OFF: do not use Supernodal Module" ON )
 
 # KLU options affecting dependencies
 option ( KLU_USE_CHOLMOD "ON (default): use CHOLMOD in KLU.  OFF: do not use CHOLMOD in KLU" ON )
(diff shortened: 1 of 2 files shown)


mmuetzel:
For sake of completeness:
I considered a different option of only requiring the detection of BLAS libraries by SuiteSparse_config if other sub-projects that require BLAS are built. But there are users/distributors that are building the SuiteSparse libraries incrementally (and package them separately). 
If these users/distributors start, e.g., by building SuiteSparse_config without any other library, that logic would have de-activated BLAS for SuiteSparse_config automatically. If they'd later try to build a library that does depend on BLAS with the version of SuiteSparse_config that they just build. That might work, but it might not behave correctly (e.g., incorrect Fortran integer size).

So, I opted for requiring the detection of BLAS libraries by SuiteSparse_config by default (independent of which other SuiteSparse sub-projects are selected). Only if a user choses *explicitly* to not require BLAS for SuiteSparse_config, it will skip that detection attempt.

DrTimothyAldenDavis:
> So, I opted for requiring the detection of BLAS libraries by SuiteSparse_config by default (independent of which other SuiteSparse sub-projects are selected). Only if a user choses _explicitly_ to not require BLAS for SuiteSparse_config, it will skip that detection attempt.

Excellent choice -- yes, I agree that's the best way to do it.

PR #946, opened 2025-05-13
Move static targets into separate exports to allow splitting them off…

opoplawski:
… in packaging

This is to start addressing the issue raised in #923.  This only touches a few libs at the moment, would need to be done for all.  Also, just making the _static export include OPTIONAL is a fairly crude heuristic, but it at least fits for the Fedora packaging scheme.   Another suggestion was to have STATIC and SHARED components, but that seems like unneeded complexity.


DrTimothyAldenDavis:
Thanks for the update.  What does this change look like for those not affected by the issue in #923? 

Also -- my organization has revised its Contributor Agreement (see https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/stable/Contributor_License/SuiteSparse%20Individual%20Contributor%20License%20Agreement%20(20241011).pdf ).  Can you sign this revised one?

mmuetzel:
~~If I understand the proposed changes correctly, users would need to write something like `find_package(AMD_static)` in their projects to import the targets for the static libraries.~~  (Edit: I misunderstood that part. Users will still use `find_package(AMD)` after this change. The next part is still an issue - I think.)
If they are trying to link to SuiteSparse libraries that had been built with `-DBUILD_STATIC_LIBS=ON -DBUILD_SHARED_LIBS=OFF`, `find_package(AMD)` would result in an error (saying that no AMD package is installed) even though (static) SuiteSparse libraries *are* installed.

I don't know if I like this approach. (Edit 2: But maybe it is possible to address that issue quite easily. E.g., by importing both target files with `OPTIONAL`...)

But maybe, I'm misreading what is actually done.

opoplawski:
The goal is to make this transparent to users, but to do that we do need to get the logic right.  Making both imports OPTIONAL might be the simplest, if not exactly the most correct.  Maybe we make these dependent on whether BUILD_STATIC_LIBS or BUILD_SHARED_LIBS was set?

I was looking for ways to determine whether or not the build doing the find_package() call is doing a static or shared build, but I'm not sure if there is a good way to do that or not.

opoplawski:
This version seems more promising so far with the checks in my fork though they are not yet complete.  Thoughts?

opoplawski:
Things are green here - https://github.com/opoplawski/SuiteSparse/actions  If this approach is acceptable, I can make the changes for all of the suitesparse libraries.

DrTimothyAldenDavis:
I just noticed that the pull request is targeting the dev branch.  Can you change it to use the dev2 branch?  That has the latest code-in-progress.

DrTimothyAldenDavis:
The dev and dev2 branches are almost identical just now, so it should be pretty simple.  They differ only in  the .github/workflow files so there shouldn't by any conflict.

opoplawski:
Changed to dev2

mmuetzel:
I haven't tested them yet. But the proposed changes look good to me. 👍

The only comment is that the convention for the CMake files in this project is to intend bodies of if-conditions by four spaces.

This pattern can probably be used for all subprojects.

mmuetzel:
Thank you for the update. This looks good to me. (I haven't tested locally yet though.)

@DrTimothyAldenDavis: Could you please approve running the workflows of this PR?

mmuetzel:
There seems to be an issue in the generation of the `KLU_CHOLMODTargets_static.cmake` file:
```
-- KLU version: 2.3.5
  -- KLU include: /home/runner/work/SuiteSparse/SuiteSparse/include/suitesparse
  -- KLU library: /home/runner/work/SuiteSparse/SuiteSparse/lib/libklu.so.2.3.5
  -- KLU static:  /home/runner/work/SuiteSparse/SuiteSparse/lib/libklu.a
  CMake Error at /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/KLU_CHOLMOD/KLU_CHOLMODConfig.cmake:106 (include):
    include could not find requested file:
  
      /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/KLU_CHOLMOD/KLU_CHOLMODTargets_static.cmake
    CMakeLists.txt:[100](https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/15090132122/job/42444891837?pr=946#step:14:101) (find_package)
  
  
  -- KLU_CHOLMOD version: 2.3.5
  -- KLU_CHOLMOD include: /home/runner/work/SuiteSparse/SuiteSparse/include/suitesparse
  -- KLU_CHOLMOD library: /home/runner/work/SuiteSparse/SuiteSparse/lib/libklu_cholmod.so.2.3.5
  -- KLU_CHOLMOD static:  /home/runner/work/SuiteSparse/SuiteSparse/lib/libklu_cholmod.a
```

The rules to generate that file need to be added probably around here:
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/ab0faee5ec817f617e19fa991f03e86f9adf2039/KLU/CMakeLists.txt#L431-L454

mmuetzel:
There also seems to be an issue with the `LAGraphConfig.cmake` file for the case where only static libraries are built:
https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/15090132122/job/42444891841?pr=946#step:14:158
```
CMake Error at /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/LAGraph/LAGraphConfig.cmake:92 (include):
    include could not find requested file:
  
      /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/LAGraph/LAGraphTargets.cmake
    CMakeLists.txt:102 (find_package)
  
  
  CMake Error at /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/LAGraph/LAGraphConfig.cmake:127 (add_library):
    add_library cannot create ALIAS target "SuiteSparse::LAGraph" because
    target "SuiteSparse::LAGraph_static" does not already exist.
    CMakeLists.txt:102 (find_package)
  
  
  CMake Error at /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/LAGraph/LAGraphConfig.cmake:131 (add_library):
    add_library cannot create ALIAS target "SuiteSparse::LAGraphX" because
    target "SuiteSparse::LAGraphX_static" does not already exist.
    CMakeLists.txt:102 (find_package)
  
  
  CMake Error at /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/LAGraph/LAGraphConfig.cmake:134 (get_target_property):
    get_target_property() called with non-existent target
    "SuiteSparse::LAGraph".
    CMakeLists.txt:102 (find_package)
  
  
  CMake Error at /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/LAGraph/LAGraphConfig.cmake:139 (get_target_property):
    get_target_property() called with non-existent target
    "SuiteSparse::LAGraph".
    CMakeLists.txt:102 (find_package)
  
  
  CMake Error at /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/LAGraph/LAGraphConfig.cmake:141 (get_target_property):
    get_target_property() called with non-existent target
    "SuiteSparse::LAGraph".
    CMakeLists.txt:102 (find_package)
  
  
  CMake Error at /home/runner/work/SuiteSparse/SuiteSparse/lib/cmake/LAGraph/LAGraphConfig.cmake:[203](https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/15090132122/job/42444891841?pr=946#step:14:204) (suitesparse_check_exist):
    suitesparse_check_exist Macro invoked with incorrect arguments for macro
    named: suitesparse_check_exist
    CMakeLists.txt:102 (find_package)
```

That is likely because this PR doesn't (yet) change `LAGraph/config/LAGraphConfig.cmake.in`.

mmuetzel:
Thank you again for the update.

The changes are looking good to me now. 👍

The currently failing CI tests are unrelated afaict. This can be merged as it is imho.

DrTimothyAldenDavis:
OK ... here goes.  I'll need to revise the version numbers and dates of all these packages (AMD 3.3.3 will become 3.3.4, etc), after I merge this in.

PR #949, opened 2025-05-16
Replacing emulation with cross-compilation in the build-arch-emu.yaml workflow to speed up SuiteSparse builds

MigunovaAnastasia1:
# SuiteSparse cross-compilation

## Supported architectures

All architectures for which SuiteSparse cross-dependencies exist in the Ubuntu repository are supported. Namely: 
* amd64
* armhf
* ppc64el
* s390x
* riscv64

## Results of replacing emulation with cross-compilation

Replacing emulation with cross-compilation gave a significant speedup in building SuiteSparse libraries for different architectures. The build is so fast that it is now possible to build absolutely all SuiteSparse libraries, including GraphBLAS and LAGraph, on different architectures. Already including GraphBLAS and LAGraph, it now takes only about 30 minutes to run build-arch-emu.yaml (now called 'cross-compile.yaml') workflow.

@DrTimothyAldenDavis also, the ability to build and test GraphBLAS and LAGraph may help you to solve the  RISCV testing problem for the following PR DrTimothyAldenDavis/GraphBLAS#381.

### Problem
By running the cross-compile.yaml workflow it was found that testing of LAGraph on **armhf** and **riscv64** architectures works correctly, but on the rest architectures **ppc64el** and **s390x** fails (see the result of the cross-compile.yaml [run](https://github.com/MigunovaAnastasia1/SuiteSparse-cross-compile/actions/runs/15065884178)).

### Localisation of the problem 
The identified problem was localised by a binary search (using the `git bisect` command) and found in the commit 1fd4ef8.

### DIFF
--- a/.github/workflows/cross-compile.yaml
+++ b/.github/workflows/cross-compile.yaml
@@ -1,0 +1,321 @@
+name: cross-compile
+on:
+  workflow_dispatch:
+  push:
+    branches-ignore:
+      - '**/dev2'
+      - '**/*dev2'
+  pull_request:
+
+concurrency: ci-cross-compile-${{ github.ref }}
+
+env:
+  # string with name of libraries to be built
+  BUILD_LIBS: "SuiteSparse_config:Mongoose:AMD:BTF:CAMD:CCOLAMD:COLAMD:CHOLMOD:CSparse:CXSparse:LDL:KLU:UMFPACK:ParU:RBio:SPQR:SPEX:GraphBLAS:LAGraph"
+  # string with name of libraries to be checked
+  CHECK_LIBS: "SuiteSparse_config:Mongoose:AMD:BTF:CAMD:CCOLAMD:COLAMD:CHOLMOD:CSparse:CXSparse:LDL:KLU:UMFPACK:ParU:RBio:SPQR:SPEX:GraphBLAS:LAGraph"
+  # string with name of libraries that are installed
+  INSTALLED_LIBS: "SuiteSparse_config:Mongoose:AMD:BTF:CAMD:CCOLAMD:COLAMD:CHOLMOD:CXSparse:LDL:KLU:UMFPACK:ParU:RBio:SPQR:SPEX:GraphBLAS:LAGraph"
+
+
+jobs:
+
+  cross-compile:
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        # Use bash shell as default
+        shell: bash
+
+    strategy:
+      # Allow other runners in the matrix to continue if some fail
+      fail-fast: false
+
+      matrix:
+        arch: [amd64, armhf, ppc64el, s390x, riscv64]
+        include:
+          - arch: amd64
+            cc: gcc
+            ccache-max: 64M
+          - arch: armhf
+            target_system: arm-linux-gnueabihf
+            cross-toolchain: g++-9-arm-linux-gnueabihf
+            fortran-cross-toolchain: gfortran-10-arm-linux-gnueabihf
+            cc: "arm-linux-gnueabihf-gcc-9"
+            cxx: "arm-linux-gnueabihf-g++-9"
+            fc: "arm-linux-gnueabihf-gfortran-10"
+            ccache-max: 64M
+          - arch: ppc64el
+            target_system: powerpc64le-linux-gnu
+            cross-toolchain: g++-9-powerpc64le-linux-gnu 
+            fortran-cross-toolchain: gfortran-10-powerpc64le-linux-gnu
+            cc: "powerpc64le-linux-gnu-gcc-9"
+            cxx: "powerpc64le-linux-gnu-g++-9"
+            fc: "powerpc64le-linux-gnu-gfortran-10"
+            ccache-max: 64M
+          - arch: s390x
+            target_system: s390x-linux-gnu
+            cross-toolchain: g++-9-s390x-linux-gnu 
+            fortran-cross-toolchain: gfortran-10-s390x-linux-gnu
+            cc: "s390x-linux-gnu-gcc-9"
+            cxx: "s390x-linux-gnu-g++-9"
+            fc: "s390x-linux-gnu-gfortran-10"
+            ccache-max: 64M
+          - arch: riscv64
+            target_system: riscv64-linux-gnu
+            cross-toolchain: g++-9-riscv64-linux-gnu 
+            fortran-cross-toolchain: gfortran-10-riscv64-linux-gnu
+            cc: "riscv64-linux-gnu-gcc-9"
+            cxx: "riscv64-linux-gnu-g++-9"
+            fc: "riscv64-linux-gnu-gfortran-10"
+            ccache-max: 64M
+
+    name: cross-compile (${{ matrix.arch }})
+
+    env:
+      CC: ${{ matrix.cc }}
+      CXX: ${{ matrix.cxx }}
+      FC: ${{ matrix.fc }}
+      LD_LIBRARY_PATH: "/usr/${{ matrix.target_system }}/lib"
+      LIBRARY_PATH: "/usr/${{ matrix.target_system }}/lib"
+
+    steps:
+      - name: get CPU information
+        run: lscpu
+
+      - name: install some aux utilities
+        run: |
+          echo "::group::Install cross-toolchain for ${{ matrix.arch }}"
+          sudo apt -y install \
+            cmake \
+            make \
+            ccache \
+            automake \
+            libtool \
+            qemu-user-binfmt
+          echo "::endgroup::"
+
+      - name: checkout repository
+        uses: actions/checkout@v4
+
+      - name: install cross-toolchain
+        run: |
+          if [[ ${{ matrix.arch }} != 'amd64' ]]; then
+            echo "::group::Install cross-toolchain for ${{ matrix.arch }}"
+            sudo apt -y install ${{ matrix.cross-toolchain }} ${{ matrix.fortran-cross-toolchain }}
+            echo "::endgroup::"
+          else
+            echo "Target architecture is ${{ matrix.arch }}, cross-compilation isn't required, cross-toolchain installation isn't required."
+          fi
+
+      - name: configure Multiarch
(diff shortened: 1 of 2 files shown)


mmuetzel:
Nice! 👍

Maybe, it would make sense to keep one or two of the non-emulated Alpine Linux runners to keep testing against MUSL (compared to glibc on Ubuntu)...

vkutuev:
> Maybe, it would make sense to keep one or two of the non-emulated Alpine Linux runners to keep testing against MUSL (compared to glibc on Ubuntu)...

What is the best option for integrating the build and testing on Alpine Linux into the same workflow file or creating a separate one?

mmuetzel:
It might be best to keep the existing Alpine Linux file (and strip the emulated runners).
Then, add a new .yaml file with the rules for the cross-builds on Ubuntu.

MigunovaAnastasia1:
> It might be best to keep the existing Alpine Linux file (and strip the emulated runners). Then, add a new .yaml file with the rules for the cross-builds on Ubuntu.

Done.
I kept armv7, x86 and loongarch64 architectures in your build-arch-emu workflow. The last one is kept, as my cross-compilation based workflow doesn't  support it.

mmuetzel:
Thank you. That looks cleaner to me now.

Just asking: Why is there a configuration for `amd64` in the build matrix? Is it different from the existing "native" CI runners on Ubuntu?

@DrTimothyAldenDavis: Could you please approve running the workflows of this PR?

MigunovaAnastasia1:
> Just asking: Why is there a configuration for `amd64` in the build matrix? Is it different from the existing "native" CI runners on Ubuntu?

Apologies, just forgot to remove amd64 from the build matrix and yes, there are no differences from existing ‘native’ CI runners on Ubuntu.

DrTimothyAldenDavis:
Thanks for the updates.  

can you sign the SuiteSparse Contributor Agreement and send it to me at drtimothyaldendavis@gmail.com?

MigunovaAnastasia1:
> can you sign the SuiteSparse Contributor Agreement and send it to me at [drtimothyaldendavis@gmail.com](mailto:drtimothyaldendavis@gmail.com)?

I apologise for forgetting to send the agreement. I did.

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,310 @@
...
+  # string with name of libraries to be checked
+  CHECK_LIBS: "SuiteSparse_config:Mongoose:AMD:BTF:CAMD:CCOLAMD:COLAMD:CHOLMOD:CSparse:CXSparse:LDL:KLU:UMFPACK:ParU:RBio:SPQR:SPEX:GraphBLAS:LAGraph"
+  # string with name of libraries that are installed
+  INSTALLED_LIBS: "SuiteSparse_config:Mongoose:AMD:BTF:CAMD:CCOLAMD:COLAMD:CHOLMOD:CXSparse:LDL:KLU:UMFPACK:ParU:RBio:SPQR:SPEX:GraphBLAS:LAGraph"
+
+
+jobs:
+
+  cross-compile:
+    runs-on: ubuntu-latest
```
mmuetzel, line 24: Since you are adding the Ubuntu repositories for Noble Nombat, would it make sense to pin this to `runs-on: ubuntu-24.04`?

DrTimothyAldenDavis:
> The identified problem was localised by a binary search (using the git bisect command) and found in the commit https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/1fd4ef8c95b83862c10c2e04df4b58cfbfccc3dc. 


Correction:  it's a gcc bug.   No need to split the yaml file.  See the details in my comments below.

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,310 @@
...
+      matrix:
+        arch: [armhf, ppc64el, s390x, riscv64]
+        include:
+          - arch: armhf
+            target_system: arm-linux-gnueabihf
+            cross-toolchain: g++-9-arm-linux-gnueabihf
+            fortran-cross-toolchain: gfortran-10-arm-linux-gnueabihf
+            cc: "arm-linux-gnueabihf-gcc-9"
+            cxx: "arm-linux-gnueabihf-g++-9"
+            fc: "arm-linux-gnueabihf-gfortran-10"
```
mmuetzel, line 44: Just out of interest: Is there a reason why you pin the compiler to these (old) versions here?
And similarly for the other host targets.

DrTimothyAldenDavis:
I suspect the bug is occurring because the ppc architecture triggers a bug in gcc. I just found this a week ago and haven't updated Suitesparse with this change:

https://github.com/DrTimothyAldenDavis/GraphBLAS/commit/15a92919662539c89a35291aaca7f95059ab3ab6

The problem is a failed OpenMP atomic in the symbolic phase of mxm. 

So I changed the GraphBLAS cmake so that it will refuse to compile on the Power processor with gcc.   Everything works fine with clang

DrTimothyAldenDavis:
See also https://github.com/DrTimothyAldenDavis/GraphBLAS/blob/92bae92bea60c8ccaae4a8c0c939135f205e2036/Source/mxm/factory/GB_AxB_saxpy3_symbolic_template.c#L221-L233 

I've just copied over the latest dev2 branch of GraphBLAS into SuiteSparse.  This will cause the CI to fail early:  in the cmake when GraphBLAS is compiled.  The cmake script will report a failure if gcc is used to compile GraphBLAS on the Power processor.

Can the CI be changed to use clang or xlc instead of gcc, for the cross compilation?

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,310 @@
...
+            ccache-max: 64M
+          - arch: ppc64el
+            target_system: powerpc64le-linux-gnu
+            cross-toolchain: g++-9-powerpc64le-linux-gnu
```
DrTimothyAldenDavis, line 48: gcc has a bug that causes GraphBLAS to fail on the Power processor.  Is there a clang toolchain that can be used instead?

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,310 @@
...
+            target_system: powerpc64le-linux-gnu
+            cross-toolchain: g++-9-powerpc64le-linux-gnu 
+            fortran-cross-toolchain: gfortran-10-powerpc64le-linux-gnu
+            cc: "powerpc64le-linux-gnu-gcc-9"
+            cxx: "powerpc64le-linux-gnu-g++-9"
+            fc: "powerpc64le-linux-gnu-gfortran-10"
+            ccache-max: 64M
+          - arch: s390x
+            target_system: s390x-linux-gnu
+            cross-toolchain: g++-9-s390x-linux-gnu
```
DrTimothyAldenDavis, line 56: I haven't tried gcc on the s390x with GraphBLAS, but I'm guessing it has the same bug as it does on the Power CPU.  Can you try a clang toolchain instead?

DrTimothyAldenDavis:
I have a Power8-based server, which is how I found this gcc bug.  I hacked my cmake script in the GraphBLAS/dev2 branch to enable gcc, and it exactly replicates the errors seen here in LAGraph.  The errors are the same on the s390x so I'm guessing it's the same gcc bug.  So replacing gcc with clang for both the powerpc and s390x builds will likely fix the issue (or provide a workaround for the gcc bug, to be more precise).

It's not common to find a bug in a compiler but I'm fairly certain that's what's happening here.

.github/workflows/cross-compile.yaml
mmuetzel: Untested: It might be possible to install the `clang` package (additionally to the GCC cross toolchain) and set `CC="clang -target ${{ matrix.target_system }}"` and `CXX="clang++ -target ${{ matrix.target_system }}"`.
The easiest way to do that is probably by changing the lines that start with `cc:` and `cxx:` in the `include` section of the `matrix` definition.

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,310 @@
...
+      # Allow other runners in the matrix to continue if some fail
+      fail-fast: false
+
+      matrix:
+        arch: [armhf, ppc64el, s390x, riscv64]
+        include:
+          - arch: armhf
+            target_system: arm-linux-gnueabihf
+            cross-toolchain: g++-9-arm-linux-gnueabihf
+            fortran-cross-toolchain: gfortran-10-arm-linux-gnueabihf
```
mmuetzel, line 41: Afaict, the value of `fortran-cross-toolchain` is only used where the value of `cross-toolchain` is also used.
Would it make sense to simplify the matrix slightly by combining the two?
``` suggestion
            cross-toolchain: g++-9-arm-linux-gnueabihf gfortran-10-arm-linux-gnueabihf
```

And similarly for the other host targets.

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,310 @@
...
+
+    strategy:
+      # Allow other runners in the matrix to continue if some fail
+      fail-fast: false
+
+      matrix:
+        arch: [armhf, ppc64el, s390x, riscv64]
+        include:
+          - arch: armhf
+            target_system: arm-linux-gnueabihf
```
mmuetzel, line 39: Maybe, rename `target_system` to `target-triple`. Imho, that would make it easier to understand what it is without having to look at its value.

DrTimothyAldenDavis:
Oops!  I meant to authorize the workflow run but I clicked on "merge" instead.  I'd like to unmerge it, not because I don't want to accept it (I do), but because it's still in progress and I'd rather wait until all the tests pass.  Sorry for any confusion.  I'll try reverting the merge ... hopefully I won't break anything the process ...

DrTimothyAldenDavis:
Ack ... I thought that "revert" would place this pull request back into the "open" state, but it didn't.  So I broke it after all. 
 Sorry!

Can you reopen the pull request again?

MigunovaAnastasia1:
> Ack ... I thought that "revert" would place this pull request back into the "open" state, but it didn't. So I broke it after all. Sorry!
> 
> Can you reopen the pull request again?

Sure, no problem!

PR #952, opened 2025-05-22
LAGraph: Avoid GCC extensions to the standards

mmuetzel:
Some compilers (notably MSVC and Clang in MSVC mode) don't support GCC extensions to the C and OpenMP standards.

Use standard compliant alternatives where MSVC fails to build currently.

Also, do not link with `libm` on platforms that do not need (e.g., MUSL, MinGW) or do not have (e.g., MSVC) that library.


mmuetzel:
Apparently, one test for LAGraph is failing in some configurations with an error like the following:
```
      Start 70: LAGraphX_peer_pressure
  1/1 Test #70: LAGraphX_peer_pressure ...........***Failed    0.13 sec
  Test peer_pressure...                             Test interrupted by SIGSEGV.
  Test peer_pressure_errors...                    
  result: -2 LAGraph assertion "c_f != ((void*)0)" failed (file /home/runner/work/SuiteSparse/SuiteSparse/LAGraph/experimental/algorithm/LAGr_PeerPressureClustering.c, line 101): status: -2
  
  result: -1003 LAGraph failure (file /home/runner/work/SuiteSparse/SuiteSparse/LAGraph/experimental/algorithm/LAGr_PeerPressureClustering.c, line 113): G->AT is required
  
  result: -1003 LAGraph failure (file /home/runner/work/SuiteSparse/SuiteSparse/LAGraph/experimental/algorithm/LAGr_PeerPressureClustering.c, line 113): G->AT is required
```
Afaict, all platforms where that fails have a pointer size of 32 bits (i686 or 32-bit ARM).

That test already failed before the changes from this PR on these platforms. So, it is not caused by the changes from here.

Maybe, a mismatch in integer size for arrays that are used as indices to arrays somewhere?

DrTimothyAldenDavis:
Thanks for the update.  I'll merge it in and then see if I can track down the segfault.

DrTimothyAldenDavis:
We might also have trouble with the C11 _Generic keyword in GraphBLAS, which is used by LAGraph.  If the Microsoft cl compiler is used, then _Generics don't work and this would cause LAGraph to fail to be compiled.  See https://github.com/GraphBLAS/LAGraph/issues/323 which I just posted yesterday.  GraphBLAS itself will still compile since it doesn't use _Generic itself.  It just makes them available in its GraphBLAS.h header.

It's possible that this issue could be encountered when extending the SuiteSparse github CI to more compilers and platforms.

DrTimothyAldenDavis:
The test_peer_pressure_errors tests is supposed to produce those error messages; that's a successful test.  It's the test_peer_pressure segfault that is the problem.

I found the bug in LAGr_PeerPressureClustering:  https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/25a81c02921f7eea7bf64dd4d11e95b985546b14

The tests are running now, but that would be the likely cause of the segfault.

PR #954, opened 2025-05-22
CI (MSVC): Build more configurations with `clang-cl`

mmuetzel:
Running the tests for LAGraph takes pretty long (approx. 30 minutes) if a compiler (like `cl` from MSVC) is used that only supports OpenMP 2.0.

Build more configurations for the MSVC target with LLVM `clang-cl` which supports OpenMP 5.1.

This depends on the changes from #952 which is needed to build LAGraph for a MSVC target.

### DIFF
--- a/.github/workflows/root-cmakelists-msvc.yaml
+++ b/.github/workflows/root-cmakelists-msvc.yaml
@@ -38,34 +38,34 @@ jobs:
         openmp: [with]
         cuda: [without]
         link: [both]
-        cc: [cl]
-        cxx: [cl]
+        cc: [clang-cl]
+        cxx: [clang-cl]
         include:
           - openmp: with
             cuda: with
             link: both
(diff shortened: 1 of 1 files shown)


DrTimothyAldenDavis:
Is there a simple way to run this on my revised dev2 branch?  I've now fixed some of the test failures, such as the LAGraph segfault.

mmuetzel:
I've rebased the PR on the current HEAD of the dev2 branch. And the runners that previously failed are green now. 👍

DrTimothyAldenDavis:
I can merge this in now, but let me know when it's ready -- you had tagged it as a Draft so I don't want to merge it in without asking you first.

mmuetzel:
It's ready to be merged now. I just forgot to convert it to "ready to review".
But it looks like you did that already. 👍

PR #955, opened 2025-05-22
Replacing emulation with cross-compilation in the build-arch-emu.yaml workflow to speed up SuiteSparse builds

MigunovaAnastasia1:
# SuiteSparse cross-compilation

## Supported architectures

All architectures for which SuiteSparse cross-dependencies exist in the Ubuntu repository are supported. Namely: 
* amd64
* armhf
* ppc64el
* s390x
* riscv64

## Results of replacing emulation with cross-compilation

Replacing emulation with cross-compilation gave a significant speedup in building SuiteSparse libraries for different architectures. The build is so fast that it is now possible to build absolutely all SuiteSparse libraries, including GraphBLAS and LAGraph, on different architectures. Already including GraphBLAS and LAGraph, it now takes only about 30 minutes to run build-arch-emu.yaml (now called 'cross-compile.yaml') workflow.

@DrTimothyAldenDavis also, the ability to build and test GraphBLAS and LAGraph may help you to solve the  RISCV testing problem for the following PR DrTimothyAldenDavis/GraphBLAS#381.

### Problem
By running the cross-compile.yaml workflow it was found that testing of LAGraph on **armhf** and **riscv64** architectures works correctly, but on the rest architectures **ppc64el** and **s390x** fails (see the result of the cross-compile.yaml [run](https://github.com/MigunovaAnastasia1/SuiteSparse-cross-compile/actions/runs/15065884178)).

### Localisation of the problem 
The identified problem was localised by a binary search (using the `git bisect` command) and found in the commit 1fd4ef8.

### DIFF
--- a/.github/workflows/cross-compile.yaml
+++ b/.github/workflows/cross-compile.yaml
@@ -52,10 +52,10 @@ jobs:
     name: cross-compile (${{ matrix.arch }})
 
     env:
-      CC: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && "clang" || "${{ matrix.target-triple }}-gcc" }}
-      CXX: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && "clang++" || "${{ matrix.target-triple }}-g++" }}
-      CFLAGS: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && "--target=${{ matrix.target-triple }}" }}
-      CXXFLAGS: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && "--target=${{ matrix.target-triple }}" }}
+      CC: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && 'clang' || ${{ matrix.target-triple }}-gcc }}
+      CXX: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && 'clang++' || ${{ matrix.target-triple }}-g++ }}
+      CFLAGS: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && --target=${{ matrix.target-triple }} }}
+      CXXFLAGS: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && --target=${{ matrix.target-triple }} }}
       FC: ${{ matrix.target-triple }}-gfortran
       LD_LIBRARY_PATH: "/usr/lib/${{ matrix.target-triple }}"
 
--- a/.github/workflows/cross-compile.yaml
+++ b/.github/workflows/cross-compile.yaml
@@ -171,8 +171,8 @@ jobs:
 
       - name: build
         run: |
-          echo "clang --version"
-          clang --version
+          echo "${CC} --version"
+          ${CC} --version
           IFS=:
           for lib in ${BUILD_LIBS}; do
             printf "   \033[0;32m==>\033[0m Building library \033[0;32m${lib}\033[0m\n"
--- a/.github/workflows/cross-compile.yaml
+++ b/.github/workflows/cross-compile.yaml
@@ -186,10 +186,8 @@ jobs:
 
       - name: build
         run: |
-          echo "${{ matrix.cc }} --version"
-          ${{ matrix.cc }} --version
-          echo "${{ matrix.cc }} -dumpmachine"
-          ${{ matrix.cc }} -dumpmachine
+          echo "clang --version"
+          clang --version
           IFS=:
           for lib in ${BUILD_LIBS}; do
             printf "   \033[0;32m==>\033[0m Building library \033[0;32m${lib}\033[0m\n"
--- a/.github/workflows/cross-compile.yaml
+++ b/.github/workflows/cross-compile.yaml
@@ -76,8 +76,7 @@ jobs:
       CFLAGS: "--target=${{ matrix.target-triple }}"
       CXXFLAGS: "--target=${{ matrix.target-triple }}"
       FC: ${{ matrix.target-triple }}-gfortran-14
-      LD_LIBRARY_PATH: "/usr/${{ matrix.target-triple }}/lib"
-      LIBRARY_PATH: "/usr/${{ matrix.target-triple }}/lib"
+      LD_LIBRARY_PATH: "/usr/lib/${{ matrix.target-triple }}"
 
     steps:
       - name: get CPU information
--- a/.github/workflows/cross-compile.yaml
+++ b/.github/workflows/cross-compile.yaml
@@ -52,10 +52,10 @@ jobs:
     name: cross-compile (${{ matrix.arch }})
 
     env:
-      CC: clang
-      CXX: clang++
-      CFLAGS: "--target=${{ matrix.target-triple }}"
-      CXXFLAGS: "--target=${{ matrix.target-triple }}"
+      CC: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && 'clang' || '${{ matrix.target-triple }}-gcc' }}
+      CXX: ${{ (matrix.arch == 'ppc64el' || matrix.arch == 's390x') && 'clang++' || '${{ matrix.target-triple }}-g++' }}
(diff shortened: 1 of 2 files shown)


MigunovaAnastasia1:
So far there are some problems with using clang as a compiler. I'll try to fix them a bit later.

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,312 @@
...
+            qemu-user-binfmt
+          echo "::endgroup::"
+
+      - name: checkout repository
+        uses: actions/checkout@v4
+
+      - name: install cross-toolchain
+        run: |
+          echo "::group::Install cross-toolchain for ${{ matrix.arch }}"
+          sudo apt -y install clang gfortran-14-${{ matrix.target-triple }}
```
mmuetzel: You still need to install the cross-compiler toolchain (additionally to `clang`).
Also, installing the current compiler version for the respective version of Ubuntu is probably fine.
``` suggestion
          sudo apt -y install clang g++-${{ matrix.target-triple }} gfortran-${{ matrix.target-triple }}
```

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,312 @@
...
+            # cc: "riscv64-linux-gnu-gcc-9"
+            # cxx: "riscv64-linux-gnu-g++-9"
+            # fc: "riscv64-linux-gnu-gfortran-14"
+            ccache-max: 64M
+
+    name: cross-compile (${{ matrix.arch }})
+
+    env:
+      CC: clang -target ${{ matrix.target-triple }}
+      CXX: clang++ -target ${{ matrix.target-triple }}
```
mmuetzel: If I understood @DrTimothyAldenDavis's comments in the previous PR correctly, he suggested to use `clang` only for the target(s) for which GCC doesn't work.

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,312 @@
...
+      FC: ${{ matrix.target-triple }}-gfortran-14
```
mmuetzel: Using the current version of gfortran of the respective version of Ubuntu is probably fine.
``` suggestion
      FC: ${{ matrix.target-triple }}-gfortran
```

.github/workflows/cross-compile.yaml
```diff
@@ -0,0 +1,312 @@
...
+        shell: bash
+
+    strategy:
+      # Allow other runners in the matrix to continue if some fail
+      fail-fast: false
+
+      matrix:
+        arch: [armhf, ppc64el, s390x, riscv64]
+        include:
+          - arch: armhf
```
mmuetzel, line 40: At some point, it would probably make sense to let the ARM target be built on an ARM build system. That would very likely be much faster. (At least when it comes to running the tests.)

See, e.g., what I recently added for the CI of Octave:
https://github.com/gnu-octave/octave/blob/default/.github/workflows/make-cross.yaml#L66-L123

That was very much inspired by what you are suggesting here. So, thank you very much for that. ♥️

.github/workflows/cross-compile.yaml
DrTimothyAldenDavis: Thanks for reposting this pull request.

Regarding clang vs gcc:  It's only on the s390 and ppc that clang must be used.  gcc has a compiler bug on those two architectures that causes GraphBLAS to be incorrectly compiled, leading to a LAGraph test failure.  For the rest, gcc can be used.

I've since revised the cmake for GraphBLAS so that it reports an error if gcc is attempted to be used to compile GraphBLAS.  That's in the latest version of my dev2 branch:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/ef510ab13c32824c1356f97a25c200f2a3c3cc4b/GraphBLAS/cmake_modules/GraphBLAS_compiler_options.cmake#L35-L37

DrTimothyAldenDavis:
I've revised my dev2 branch to fix some of the test failures on the 32-bit armv7 and mingw32 systems, so some of these errors in the CI tests should be fixed now.

DrTimothyAldenDavis:
Can you rebase this PR on the latest dev2 branch?  That will fix many of the failed github CI tests.

.github/workflows/cross-compile.yaml
MigunovaAnastasia1: When setting up clang it slipped my mind that it only needs to be configured for ppc and s390x, I'll fix that shortly

MigunovaAnastasia1:
> Can you rebase this PR on the latest dev2 branch? That will fix many of the failed github CI tests.

Sorry for the delay. Rebased.

.github/workflows/cross-compile.yaml
MigunovaAnastasia1: I'm very pleased that you were inspired by it!

MigunovaAnastasia1:
Sorry, I accidentally pushed it here, it's not ready for review yet.

DrTimothyAldenDavis:
Most or perhaps all of the github CI tests will likely pass if you rebase this with the latest dev2 branch, where many of these issues have been fixed.

mmuetzel:
Are the tests that are failing for LAGraph on s390x something that we saw before? Or are they new?
```
  The following tests FAILED:
  	 56 - LAGraphX_RichClubCoefficient (Failed)
  Errors while running CTest
  	 60 - LAGraphX_SwapEdges (Failed)
```

In case that should matter: s390x has big endian byte order (unlike most other platforms).

MigunovaAnastasia1:
> Are the tests that are failing for LAGraph on s390x something that we saw before? Or are they new?

The tests that were failing at the time of posting my PR are different from the ones you mentioned above. Here are the tests that were failing then:
```
  The following tests FAILED:
  	  2 - LAGraph_BreadthFirstSearch (Failed)
  	 45 - LAGraphX_MaximalIndependentSet (Failed)
```
But the tests you  mentioned above started failing the moment I replaced gcc with clang (on s390x)

DrTimothyAldenDavis:
The BFS and MaximalIndependentSet failures are because of a gcc bug that causes a failure in an atomic operation in GraphBLAS.  Those have been "fixed" by forcing a cmake fatal error if an attempt is made to compile GraphBLAS with gcc.

I have a Power8 system so I could track down that bug.  I don't have access to an s390 system, but the test failures were identical on the 2 systems.

I think this new failure is unrelated.  I only just recently added the RichClubCoefficient and SwapEdges method to LAGraph -- after this PR was first written, so they weren't exercised until recently.

DrTimothyAldenDavis:
Both the Power and s390x architectures are big-endian so that's not likely the issue.  It will take me some time to track down why the LAGraph tests are failing on the s390x: either an LAGraph bug in these 2 new methods (they rely on each other so that seems likely), an GraphBLAS bug, or perhaps yet another compiler or system bug.  Doing so will require me to revise LAGraph and/or GraphBLAS, post the updates here on github (say on a new branch for the purpose) and try the github CI.

I'm about ready to release SuiteSparse 7.11.0 though, and I'd rather just state, for now, that LAGraph and GraphBLAS are not supported on the s390x.  

I think the short term solution is to remove the s390x from the cross-compile tests and leave them in the build-arch-emu.yaml test (which don't test LAGraph or GraphBLAS).

If I can track down the issue quickly, I'll include the s390x in the github CI for SuiteSparse 7.11.0.  Otherwise, I'll revise the yaml files to exclude the s390x from the cross-compile.yaml, and then and release SuiteSparse 7.11.0 as is.

Either way, I can go ahead and merge this in now.

DrTimothyAldenDavis:
I found the bug in the new LAGraph routine (FastAssign) that was causing these test failures on the s390x, and a few other bugs in the new methods (RCC and SwapEdges).

@mmuetzel : you were right, it was in fact a side effect of endian-ness, and a latent bug in little endian systems too ( the bug just didn't trigger a fault).  The LAGraph_FastAssign method was passing a pointer to a bool scalar where a pointer to int32_t was expected, as a scalar output parameter of a GraphBLAS method.  Oops,  So GraphBLAS was writing 4 bytes into a 1-byte sized answer ... which happened to kill the big endian s390x but could have easily killed any algorithm on any system.  It just happened to not cause any damage on the other systems.

I'll post an update shortly.

DrTimothyAldenDavis:
All tests are now passing.  Yay!

PR #956, opened 2025-05-23
LAGraph: Fix some recently introduced build issues

mmuetzel:
See the commit messages.


mmuetzel:
One tests seems to be segfaulting on all Windows configurations.
To be honest, I'm not sure why it doesn't abort on Linux, too. Maybe, that is because trying to allocate even humungous amounts of memory never fails on Linux...

From what I can tell, the test is failing in `LAGr_SwapEdges`. That function is being called with `e` set to 104. For that number, `__builtin_clzl` correctly returns 25.
With that, `ehash_size` is calculated to 2^(67-25). That is approximately 4.4e12.
What comes next is pretty surprising to me: That huge number is used to set the number of elements of a vector.
`GRB_TRY (GrB_Vector_new(&exists, GrB_INT8, ehash_size)) ;`
If I understand correctly, that vector is set to full: `GRB_TRY(GrB_set (exists, GxB_BITMAP | GxB_FULL, GxB_SPARSITY_CONTROL)) ;`
At that point, it crashes for me on Windows. (I don't have 4 TB of free memory.)

Am I misunderstanding something? What should actually be happening here?

mmuetzel:
Anyway. These tests already failed on Windows/MinGW and on 32-bit Linux before the changes from this PR.

With the changes from here, the build issues are fixed that previously happened for Windows/MSVC and for configurations that build only static libraries.

DrTimothyAldenDavis:
Good question.  @GomezGab : can you take a look?  I think there must be a bug in computing ehash_size on Windows.  If I run test_SwapEdges on Linux (after uncommenting a printf to print the value ehash_size) the largest ehash_size I see is 524,288:

```
$ ./build/experimental/test/test_SwapEdges | grep "Hash Size" | uniq -c
      2 Hash Size: 1024
      2 Hash Size: 8192
      2 Hash Size: 16384
      2 Hash Size: 8192
      2 Hash Size: 524288
      2 Hash Size: 32768
      1 Hash Size: 32
    894 Hash Size: 1024
    725 Hash Size: 8192
    726 Hash Size: 16384
    726 Hash Size: 8192
    726 Hash Size: 1024
```

If ehash_size is computed as 4.4e12 on Windows in the test_SwapEdges, then it must be computed incorrectly.

mmuetzel:
I think I found the issue: `long` has a size of 32 bit on 32 bit systems (ILP32) and on Windows (independent on whether its pointer size is 32 bit, ILP32, or 64 bit, LLP64).
`long long` has a size of 64 bit everywhere.

I followed up by a change that uses `__builtin_clzll` instead of `__builtin_clzl`. That should be fixing it (I think).

DrTimothyAldenDavis:
Ah, that would cause Windows to fail.

For some reason, GitHub is giving me errors on reading this page and I can't see the status of the CI tests, or even whether or not the PR can be merged.  It says "Merge status cannot be loaded
Try reloading the page, or if the problem persists contact support." ... oh well...

The other updates in this PR look great -- thanks!

mmuetzel:
I'm seeing the same message. (The same for #955.) It looks like GitHub is having some issues. Their status page is (still?) green:
https://www.githubstatus.com/

But I'd guess somethings not quite working on their side currently.

The CI tests are still visible on the "Actions" tab for me: https://github.com/DrTimothyAldenDavis/SuiteSparse/actions

Edit: And they noticed: https://www.githubstatus.com/incidents/y71cytztvck9

GomezGab:
All the changes look good to me. Would it be better to have a CLZ macro or a LOG2 macro like in GraphBLAS, and then do
`int shift_e = 63 - LOG2(size);`?

GomezGab:
On second thought, why not just use `shift_e = 63 - (int) floor (log2 ((double) e));` on every platform? It should be defined on all platforms. I don't think there was really a good reason for me to use clz in the first place.

DrTimothyAldenDavis:
> On second thought, why not just use `shift_e = 63 - (int) floor (log2 ((double) e));` on every platform? It should be defined on all platforms. I don't think there was really a good reason for me to use clz in the first place.

This would probably be better.  It's more portable, and doesn't require a revised cmake to check if __builtin_clzll exists.  The function is just called once, so there's no performance benefit anyway.

DrTimothyAldenDavis:
I'll go ahead and merge this in, and we can remove the czll afterwards.

DrTimothyAldenDavis:
See https://github.com/DrTimothyAldenDavis/SuiteSparse/commit/be0eacbbc01b27b349b9a1de876e81fc3f52b3fc where I've removed all uses of __builtin_clzll.  Using floor(log2(e)) instead.  I tested this and it gets the same hash sizes as the clzll method.  Performance is not an issue since it's only used once inside the LAGr_SwapEdges algorithm anyway, and it's more portable.

PR #989, opened 2025-07-22
emit an error or warning if the C/C++/Fortran compilers do not match (keeping default settings in prior SuiteSparse releases)

DrTimothyAldenDavis:
If the Intel icx compiler is in use, and libiomp is to beused with the BLAS, and the Fortran compiler is from GNU, then SUITESPARSE_USE_FORTRAN  = ON causes a conflict.  It results in find_package ( BLAS ... ) finding the Intel MKL library, but with libgomp instead of libiomp.  But icx brings in libiomp.  The result is two OpenMP libraries, which is very bad.

Also included in this PR are revisions to the ParU benchmarking.


DrTimothyAldenDavis:
See  FindBLAS.cmake, where the Intel MKL BLAS switches from libiomp to libgomp if Fortran has been enabled:

https://github.com/Kitware/CMake/blob/8eeeeb7efe3f005c78818bf4806b80c33d68005d/Modules/FindBLAS.cmake#L490-L502

```
# BLAS in the Intel MKL 10+ library?
if(BLA_VENDOR MATCHES "Intel" OR BLA_VENDOR STREQUAL "All")
  if(NOT BLAS_LIBRARIES)
    if(CMAKE_C_COMPILER_LOADED OR CMAKE_CXX_COMPILER_LOADED)
      # System-specific settings
      if(WIN32)
... 
      else()
...
        # Switch to GNU Fortran support layer if needed (but not on Apple, where MKL does not provide it)
        if(CMAKE_Fortran_COMPILER_LOADED AND (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU" OR CMAKE_Fortran_COMPILER_ID STREQUAL "LCC") AND NOT APPLE)
            set(BLAS_mkl_INTFACE "gf")
            set(BLAS_mkl_THREADING "gnu")
            set(BLAS_mkl_OMP "gomp")
        else()
            set(BLAS_mkl_INTFACE "intel")
            set(BLAS_mkl_THREADING "intel")
            set(BLAS_mkl_OMP "iomp5")
        endif()
        set(BLAS_mkl_LM "-lm")
        set(BLAS_mkl_LDL "-ldl")
      endif()
```

Note the comment `# Switch to GNU Fortran support layer if needed `.  This causes a conflict because libiomp is loaded but the BLAS (above) asks for libgomp if GNU Fortran has been enabled. Both OpenMP libraries then get linked in all libraries of SuiteSparse, which is a serious problem.

The best solution would be to use the Intel Fortran compiler (ifort or ifx) when using the Intel MKL BLAS, but that's not always possible. The Intel Fortran compiler is not a default part of the Intel oneAPI suite.

mmuetzel:
I might be misunderstanding the motivating factors. But would it make sense to expect users with compilers that are incompatible to their BLAS implementation to manually disable `SUITESPARSE_USE_FORTRAN` instead of requiring everyone else to manually enable that flag?

mmuetzel:
If you think that switching the default is the way to go, it might make sense to manually enable that flag in *all* CI workflows (none of them use incompatible compiler combinations) to make sure the respective parts of the sources continue to be able to be built...

DrTimothyAldenDavis:
I hit this because the Intel OneAPI does not come with ifx by default.  It just has the intel icx compiler and the Intel MKL BLAS.  It doesn't complain about using the BLAS, and the BLAS works.  But the MKL BLAS then links with libgomp, while the rest of the libraries (UMFPACK, SPQR, CHOLMOD and SPQR) link against libgomp.

The build doesn't complain and everything "works" but the performance is seriously degraded.

The alternative would be to force the end user to use icx and ifx together, or use gcc and gfortran together.  But that's very restrictive.  Most users of SuiteSparse don't really need the few Fortran routines it provides.

If I were to have Fortran enabled by default, it would be possible to check this condition in my SuiteSparsePolicy.cmake:

`        if(CMAKE_Fortran_COMPILER_LOADED AND (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU" OR CMAKE_Fortran_COMPILER_ID STREQUAL "LCC") AND NOT APPLE)
`

and if that condition holds, and also the C / CXX compilers are icx and icpx, then I could error out and refuse to build SuiteSparse.  But that's rather extreme.

Most of the time, when someone downloads the Intel OneAPI by default (getting icx and the mkl, but not ifx) they want to use both icx and mkl, and they don't care about Fortran.  That's a common use case.   Forcing them to use gcc and the MKL is an option, or I can force them to download ifx as well, but both restrictions are rather extreme.  The better solution is to simply disable Fortran by default, so there's no conflict.

If the end user wants to use Fortran, they can ensure that they are using either all-Intel or all-GNU compilers and then they can turn on SUITESPARSE_USE_FORTRAN.  I've added this to the SuiteSparse/ChangeLog:

    * all packages:  Fortran now disabled by default, with the default setting
        of SUITESPARSE_USE_FORTRAN changes from ON to OFF.  If the icx compiler
        and the Intel MKL BLAS are available but ifx is not, then loading the
        GNU Fortran compiler causes the cmake FindBLAS to link the Intel MKL
        against libgomp, but at the same time cmake links the application with
        libiomp.  Including two OpenMP libraries, libiomp and libgomp, in one
        application can cause serious performance issues.  To avoid this issue,
        SUITESPARSE_USE_FORTRAN has been set to OFF.  This has little impact on
        SuiteSparse (it disables the AMD Fortran routines, and removes the
        Fortran interface to UMFPACK).  If you wish to use those features,
        re-enable the SUITESPARSE_USE_FORTRAN, but be sure to use either gcc
        and gfortran together, or icx and ifx together.  Do not mix-and-match
        the compilers.


I will also add that discussion to the SuiteSparsePolicy.cmake where I document SUITESPARSE_USE_FORTRAN.

Yes, it would make sense for the CI to enable this setting where it can be done safely.

DrTimothyAldenDavis:
I've added -DSUITESPARSE_USE_FORTRAN=ON to all CI workflows.

mmuetzel:
If I understand correctly, combining different implementations of OpenMP is always a bad idea. If you use a compiler (be it for C, C++, Fortran, or for other languages that support OpenMP) that uses one implementation of OpenMP and link to libraries that had been built with a different implementation of OpenMP, you are on your own.
Testing for any and all combinations that can be problematic, is cumbersome (or impossible). 
In my opinion, that is a user error. And it is not the responsibility of the build system to prevent the user from any potential error...

mmuetzel:
> * all packages:  Fortran now disabled by default, with the default setting
>     of SUITESPARSE_USE_FORTRAN changes from ON to OFF.  If the icx compiler
>     and the Intel MKL BLAS are available but ifx is not, then loading the
>     GNU Fortran compiler causes the cmake FindBLAS to link the Intel MKL
>     against libgomp, but at the same time cmake links the application with
>     libiomp.  Including two OpenMP libraries, libiomp and libgomp, in one
>     application can cause serious performance issues.  To avoid this issue,
>     SUITESPARSE_USE_FORTRAN has been set to OFF.  This has little impact on
>     SuiteSparse (it disables the AMD Fortran routines, and removes the
>     Fortran interface to UMFPACK).  If you wish to use those features,
>     re-enable the SUITESPARSE_USE_FORTRAN, but be sure to use either gcc
>     and gfortran together, or icx and ifx together.  Do not mix-and-match
>     the compilers.

Should that start with "Fortran **is** now disabled by default, with the default setting of `SUITESPARSE_USE_FORTRAN` **changed** from `ON` to `OFF`"? (Missing "is", and "changed" instead of "changes")

DrTimothyAldenDavis:
oops.  Yes, that would be better.  Thanks for the catch.

I'm open to other solutions to this issue.  It's just really dangerous but all-too-easy to mix icx and gfortran, and get two OpenMP libraries when using the Intel MKL.

mmuetzel:
> I'm open to other solutions to this issue.

My preferred "solution" would be to instruct users to use compatible combinations of compilers. 
And for compilers that can use different implementations of OpenMP (like Clang): Configure the compilers to use compatible implementations of OpenMP.

(That also goes for the compiler configuration that had been used to build all linked in libraries.)

DrTimothyAldenDavis:
The problem is that most users (even me) wouldn't know they are using any Fortran compiler at all; SuiteSparse is almost exclusively C/C++.  The Fortran codes in SuiteSparse are perhaps 30 years old and I doubt they are used very much.  An end user like that wouldn't give any thought to Fortran, and so they could easily ignore the warning.  They can easily build with icx and the Intel MKL BLAS without any thought of Fortran because they aren't using it.  But then it silently causes a massive performance hit.

This bug came up when I was benchmarking ParU, which doesn't use Fortran at all so I didn't give any thought to the fact I was using icx and gfortran together.

So it seems safest to me to simply disable Fortran by default.  Then if an end-user (or distro manager) want to use it for the few Fortran codes in SuiteSparse, they can turn it on, knowing the caveat I've added to the SuiteSparsePolicy.cmake discussion of SUITESPARSE_USE_FORTRAN.

DrTimothyAldenDavis:
The alternative would be to error out of SUITESPARSE_USE_FORTRAN (or HAS Fortran) is ON, but the compilers are not compatible.  The cmake build system could refuse to mix compilers from different toolchains.

The downside to that is that I'm not sure how to test for compatible compilers.  I can check icx + gfortran, or gcc + ifx, and error out if I see those paired together with SUITESPARSE_USE_FORTRAN ON.  But then I would also need to check clang + gfortran, gcc + flang, etc, and that seems to hard to add all those rules to cmake.

This might be too extreme, as well.  The end user might want to use icx + gfortran anyway, because they could be using OpenBLAS or some other BLAS, and not have this issue.  There are too many combinations to check: some are OK and some are silent disasters.

mmuetzel:
> * GNU Fortran compiler causes the cmake FindBLAS to link the Intel MKL
>   against libgomp, but at the same time cmake links the application with
>   libiomp.

That part of the changelog is confusing to me. The Intel MKL has already been built at that point. `FindBLAS` cannot "re-link" these binaries to a different OpenMP library from the one with which it was built originally, can it? So, CMake (correctly) links to libiomp if you use a version of Intel MKL that was built with that. Note that the version of Intel MKL that is distributed by Debian is built against LLVM OpenMP (i.e., libomp) not against libiomp.
I'm also not sure what you mean by "cmake links the application with libiomp".

DrTimothyAldenDavis:
I should write 'package' not 'application'.  What happens is that UMFPACK, CHOLMOD, SPQR, and ParU all link against libiomp and libgomp, together.  Not the end user application.

mmuetzel:
If I understand the case where you ran into that correctly, the following happened:
* You are using an implementation of the Intel MKL that was built with Intel's implementation of OpenMP libiomp. So, CMake (correctly) links with libiomp.
* You are using a compiler from the GCC and instruct it to use OpenMP. Afaict, compilers from the GCC can *only* use the GNU implementation of OpenMP libgomp. So, you end up linking to libgomp as well.

mmuetzel:
The updated version of the changelog is still as confusing to me as the version before that.
Maybe, my previous comment helps?

mmuetzel:
If the Fortran modules in AMD and CHOLMOD do not explicitly use OpenMP, not enabling OpenMP for Fortran could also be an option...

DrTimothyAldenDavis:
> If I understand the case where you ran into that correctly, the following happened:
> 
> * You are using an implementation of the Intel MKL that was built with Intel's implementation of OpenMP libiomp. So, CMake (correctly) links with libiomp.
> * You are using a compiler from the GCC and instruct it to use OpenMP. Afaict, compilers from the GCC can _only_ use the GNU implementation of OpenMP libgomp. So, you end up linking to libgomp as well.

sort of.  It's backwards.  I used icx and icpx from the Intel compiler suite, and my default BLAS is MKL.  Since FindBLAS detected that I also have Fortran enabled, it switched to use MKL with the gnu wrapper and libgomp (see the FindBLAS.cmake logic posted above).  That's how libgomp gets in there.   But the rest of ParU, CHOLMOD, UMFPACK, and SPQR link against libiomp because I'm compiling them with icx and icpx.

I just completely forgot about Fortran since none of my Fortran codes use the BLAS or OpenMP anyway.  So it shouldn't matter what Fortran compiler I use ... or so I thought.

The Intel OneAPI download excludes their ifx Fortran compiler by default, so this is an error waiting to happen.

mmuetzel:
> sort of. It's backwards. I used icx and icpx from the Intel compiler suite, and my default BLAS is MKL. Since FindBLAS detected that I also have Fortran enabled, it switched to use MKL with the gnu wrapper and libgomp (see the FindBLAS.cmake logic posted above).

You are right.
Apparently, they are only checking for the Fortran compiler and not for any other enabled languages. They also do not check if OpenMP was enabled for any compiler at that point.
Like I wrote previously, getting that right automatically is cumbersome (or impossible). And obviously CMake gets that wrong for your combination of compilers.

Might be worth reporting to upstream CMake. In my opinion, that logic in FindBLAS.cmake is flawed, only works in some cases and breaks in other cases.

mmuetzel:
Oof. I forgot that the MSVC workflows don't have a Fortran compiler (yet?). They might get one, once LLVM Flang is ready.
But currently, we can't set `SUITESPARSE_USE_FORTRAN=ON` for them.
If they'll have a Fortran compiler at some point in the future, we'd need to remember enabling that again...

DrTimothyAldenDavis:
So the fix is either: (1) tell the user to turn off Fortran themselves, (2) don't compile SuiteSparse with icx, or (3) if the user installs the default Intel oneAPI then they must also download the non-default ifx.

Option (2) is not good because ParU prefers the Intel MKL BLAS (it makes use of a non-standard function in the Intel MKL to improve performance, in an #ifdef).  Option (3) is a hassle if the user doesn't have root priviledge or doesn't otherwise need or want any Fortran compiler.  Option (1) is hard for an end user to decide if they need to turn off Fortran (I made this mistake myself).

What makes it more confusing is that the Intel OneAPI is designed so it can mix-and-match with GNU tools.  The Intel MKL can rely on libgomp for example.  It might also be possible for icx to use libgomp as well. 

So it seems to me that the safest thing is to set SUITESPARSE_USE_FORTRAN=OFF.

DrTimothyAldenDavis:
> `FindBLAS` cannot "re-link" these binaries to a different OpenMP library from the one with which it was built originally, can it?

Yes, it can.  The Intel MKL can be used with libgomp or libiomp, and FindBLAS can select between those two options.  It selects badly however, causing a conflict later on.  That's the if-logic in FindBLAS.cmake.  It intentionally makes the switch to the GNU threadling interface and libgomp:

        # Switch to GNU Fortran support layer if needed (but not on Apple, where MKL does not provide it)
        if(CMAKE_Fortran_COMPILER_LOADED AND (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU" OR CMAKE_Fortran_COMPILER_ID STREQUAL "LCC") AND NOT APPLE)
            set(BLAS_mkl_INTFACE "gf")
            set(BLAS_mkl_THREADING "gnu")
            set(BLAS_mkl_OMP "gomp")
        else()
            set(BLAS_mkl_INTFACE "intel")
            set(BLAS_mkl_THREADING "intel")
            set(BLAS_mkl_OMP "iomp5")
        endif()


It makes this switch even though the OpenMP libraries have already been found (libiomp) for the rest of the library.  If I try to find the BLAS first, before finding OpenMP, the same problem occurs: Intel MKL links against libgomp (which it's designed be able to do) but the remainder of my library (UMFPACK, SPQR, CHOLMOD, ParU) link against libiomp.

DrTimothyAldenDavis:
> If the Fortran modules in AMD and CHOLMOD do not explicitly use OpenMP, not enabling OpenMP for Fortran could also be an option...

I think that won't fix the flawed logic in FindBLAS.cmake.  That logic there just checks if the GNU Fortran compiler is loaded.  It doesn't ask if the GNU Fortran compiler needs to use OpenMP (it doesn't in my case).

The Fortran in SuiteSparse is used for a few routines in AMD, and a Fortran wrapper for my C interface to UMFPACK.  Those codes don't call the BLAS and they don't use OpenMP.

DrTimothyAldenDavis:
> Might be worth reporting to upstream CMake. In my opinion, that logic in FindBLAS.cmake is flawed, only works in some cases and breaks in other cases.

Yes, this seems like a bug in the cmake FindBLAS, to me.  But I'm not sure what the fix needs to be.  In my case, the fix would be "ignore Fortran, go with the previously found OpenMP library and disable OpenMP and the Intel MKL in any Fortran codes that GNU Fortran compiles".  But that might be too extreme and break other uses of findBLAS.

So my workaround is to ensure `CMAKE_Fortran_COMPILER_LOADED` is false, when compiling SuiteSparse itself.  The only way to do that is to turn `SUITESPARSE_USE_FORTRAN` to OFF.

Or, to turn `SUITESPARSE_USE_FORTRAN` on but then ensure all the compilers and libraries will agree on using the same OpenMP library.  That's hard to ensure, thus my thinking to set `SUITESPARSE_USE_FORTRAN` to OFF by default.

mmuetzel:
Or option (4): Report the issue to CMake and hope that someone will fix that in their FindBLAS.cmake.
But that won't help right now.

I'd prefer option (1). If I understand the situation on macOS correctly, most users won't have any Fortran compiler anyway. Similarly on Windows (MSVC). And the situation with Intel MKL on Linux might be completely different...

But I understand if you prefer a different option.

When it comes to "re-linking": Apparently, Intel MKL installs various versions of their libraries (with different names?) that have been built with different OpenMP implementations. See, e.g., this in FindBLAS.cmake:
```
            # mkl >= 10.3
            list(APPEND BLAS_SEARCH_LIBS
              "${BLAS_mkl_START_GROUP} mkl_blas95_${BLAS_mkl_ILP_MODE} mkl_${BLAS_mkl_INTFACE}_${BLAS_mkl_ILP_MODE} mkl_${BLAS_mkl_THREADING}_thread mkl_core ${BLAS_mkl_END_GROUP} ${BLAS_mkl_OMP}")
```

DrTimothyAldenDavis:
> When it comes to "re-linking": Apparently, Intel MKL installs various versions of their libraries (with different names?) that have been built with different OpenMP implementations.

Yes, that's what I mean.  They have an MKL library that links with an "intel_thread" option, which links the BLAS with libiomp, and another with the  "gnu_thread" option that links with libgomp. They have more options as well.

DrTimothyAldenDavis:
I will add more discussion in the SuiteSparse/ChangeLog (https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/989/commits/7b9947ce6b902e6d55121d3cff9a3b150e5b849f).  In particular, distro maintainers like yourself will want to set SUITESPARSE_USE_FORTRAN back to ON, since you can guarantee that you know what you're doing (but I can't guarantee that for myself it seems :-) .

There are 2 unfortunate side effects of this change.  The Fortran calling interface from C cannot be set automatically by cmake with the FortranCInterface.cmake provided by cmake.  Instead, I have to set it myself with SUITESPARSE_C_TO_FORTRAN.  That value is set to good defaults now, where it can pass all CI tests on all platforms.  But there might be a case where that fails on another platform/compiler.  The other change is that amd.f and amdbar.f are removed from the compiled AMD library.

So a distro maintainer will want to avoid these side effects, for sure.

DrTimothyAldenDavis:
I considered an alternative:  if  `CMAKE_C_COMPILER_ID` and `CMAKE_Fortran_COMPILER_ID` do not match, then I could turn off Fortan.  But this does not work because CMAKE_Fortran_COMPILER_ID is not set until enable_language ( Fortran ) is done.  At that point, it's not possible to disable Fortran, and the FindBLAS will break.

So that option  doesn't work.

I could instead go ahead and load Fortran if `SUITESPARSE_USE_FORTRAN=ON`.  Then, if the compiler IDs of Fortran, C, and C++ don't match, I could issue a fatal error, and inform the user they either need to use 3 compilers all with the same ID, or disable Fortran. I could then keep `SUITESPARSE_USE_FORTRAN` at its current default of ON.  

I will push this possible change to this PR and see how the CI workflows handle it.

Here's a sample of the error it can generate, if I try to use CC=icx, CXX=icpx, and FC=gfortran:

```
-- Incompatible Fortran/C/C++ compilers detected:
--     Fortran:          /usr/bin/f95
--     Fortran id:       GNU
--     C                 /home/davis/intel/oneapi/2025.0/bin/icx
--     C       id:       IntelLLVM
--     C++               /home/davis/intel/oneapi/2025.0/bin/icpx
--     C++     id:       IntelLLVM
CMake Error at cmake_modules/SuiteSparsePolicy.cmake:323 (message):
  Error: Using Fortran with SuiteSparse requires that it has the same
  compiler ID as the C/C++ compilers.  Use a compatible Fortran compiler, or
  set SUITESPARSE_USE_FORTRAN to OFF.
  CMakeLists.txt:46 (include)
```
@mmuetzel : what do you think of this option?  Causing a cmake fatal error is a bit dramatic, but it would enable me to keep the default SUITESPARSE_USE_FORTRAN=ON.  A distro maintainer would surely be able to rely on a suite of 3 compilers that have matching ID's wouldn't they?  And if they don't they could simply set SUITESPARSE_USE_FORTRAN=OFF themselves.  They likely do that anyway, if they know they don't have a Fortran compiler.

DrTimothyAldenDavis:
Oh ... well I guess the CI answers my questions.  mingw with clang fails for example, with this change:

https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/16481584871/job/46597018025?pr=989

```
> CMake Error at cmake_modules/SuiteSparsePolicy.cmake:323 (message):
>     Error: Using Fortran with SuiteSparse requires that it has the same
>     compiler ID as the C/C++ compilers.  Use a compatible Fortran compiler, or
>     set SUITESPARSE_USE_FORTRAN to OFF.
>   Call Stack (most recent call first):
>     CMakeLists.txt:46 (include)
>   
>   -- Incompatible Fortran/C/C++ compilers detected:
>   --     Fortran:          C:/msys[64](https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/16481584871/job/46597018025?pr=989#step:8:65)/clang64/bin/flang.exe
>   --     Fortran id:       LLVMFlang
>   --     C                 C:/msys64/clang64/bin/cc.exe
>   --     C       id:       Clang
>   --     C++               C:/msys64/clang64/bin/c++.exe
>   --     C++     id:       Clang
>   -- Configuring incomplete, errors occurred!
```

So this test is too extreme, at least for some platforms.

DrTimothyAldenDavis:
Some of the failing CI workflows look suspicious to me.  Three are mixing clang C/C++ with GNU fortran (the two ubuntu latest, and the cross-compile for ppc4).  That will give different OpenMP libraries, if the Fortran uses OpenMP.  No SuiteSparse Fortran code uses OpenMP, but what would the BLAS use?

The mingw cases that fail look OK to me because the flang ID is LLVMFlang, and C/C++ have the ID "Clang".  That should be OK but my FATAL_ERROR is too extreme in this case.

I suppose one option would be to leave the test but change it from FATAL_ERROR to WARNING, as in the next attempt ...

DrTimothyAldenDavis:
Here is a list of the failures, if I always use FATAL_ERROR:

[build / ubuntu-latest (clang with CUDA with OpenMP, both) (pull_request)](https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/16481584874/job/46597018124?pr=989)
build / ubuntu-latest (clang with CUDA with OpenMP, both) (pull_request)Failing after 3m
[build-mingw / mingw (CLANG64) (pull_request)](https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/16481584871/job/46597018025?pr=989)
build-mingw / mingw (CLANG64) (pull_request)Failing after 4m
[build-mingw / mingw (CLANGARM64) (pull_request)](https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/16481584871/job/46597018039?pr=989)
build-mingw / mingw (CLANGARM64) (pull_request)Failing after 5m
[cross-compile / cross-compile (ppc64el) (pull_request)](https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/16481584875/job/46597018176?pr=989)
cross-compile / cross-compile (ppc64el) (pull_request)Failing after 1m
[root-cmakelists / ubuntu-latest (clang with CUDA, both) (pull_request)](https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/16481584880/job/46597018151?pr=989)
root-cmakelists / ubuntu-latest (clang with CUDA, both) (pull_request)Failing after 3m

I'll try changing this to a WARNING, unless the C compiler ID is IntelLLVM, where I will report a FATAL_ERROR.  If the C compiler is icx, libiomp will be linked, but if Fortran is GNU, then the BLAS will use libgomp, which is the nasty case I'm trying to avoid.

I don't think any of the CI workflows use the Intel compiler suite.

popsection:
Hopefully I haven't missed a detail, but using [MKLConfig](https://github.com/uxlfoundation/oneMath/blob/develop/cmake/mkl/MKLConfig.cmake) instead of the built-in FindBLAS module should solve this problem.

DrTimothyAldenDavis:
This CI now passes with the change from FATAL_ERROR to WARNING.

[build / ubuntu-latest (clang with CUDA with OpenMP, both) (pull_request)](https://github.com/DrTimothyAldenDavis/SuiteSparse/actions/runs/16483401944/job/46603021014?pr=989)
build / ubuntu-latest (clang with CUDA with OpenMP, both) (pull_request)Successful in 11m

even though it uses clang (and libomp) with GNU Fortran (which has the risk of linking with libgomp).  It uses the libopenblas.so, which I think is compiled with the GNU compiler suite but not OpenMP (I have that library on my own Ubuntu system so I can check it).

The warning gets issued, so the installer should be careful to check this case.  It does look safe because the BLAS is not using any OpenMP library at all.

DrTimothyAldenDavis:
> Hopefully I haven't missed a detail, but using [MKLConfig](https://github.com/uxlfoundation/oneMath/blob/develop/cmake/mkl/MKLConfig.cmake) instead of the built-in FindBLAS module should solve this problem.

Great suggestion -- thanks.  I would hope it would do the right thing, but it might be difficult to coordinate that with FindBLAS.  My cmake scripts port everywhere, and if I try using MKLConfig, most environments won't have it.  Is there a way to check if MKLConfig is available, and use it instead of "find_package (BLAS ...)" ?

I use the find_package ( BLAS ...) only in this cmake script, for all of my packages in SuiteSparse:  
https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/dev/SuiteSparse_config/cmake_modules/SuiteSparseBLAS.cmake

popsection:
I was imagining a solution akin to [this answer](https://gitlab.kitware.com/cmake/cmake/-/issues/25426#note_1447667) on the CMake issue tracker.

```CMake
# -- find BLAS, preferring MKL
find_package(MKL CONFIG HINTS $ENV{MKLROOT})

if(MKL_FOUND)
  get_property(BLAS_INCLUDE_DIRS TARGET MKL::MKL PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
  get_property(BLAS_LIBRARIES TARGET MKL::MKL PROPERTY INTERFACE_LINK_LIBRARIES)
else()
  find_package(BLAS)
endif()
```

I must apologize for my short/half-baked answers. I haven't tested it out yet and I've only used Intel's toolchain a few times, so I'm unsure what the best solution will be. However, I plan to work on this more tomorrow.

DrTimothyAldenDavis:
> I was imagining a solution akin to [this answer](https://gitlab.kitware.com/cmake/cmake/-/issues/25426#note_1447667) on the CMake issue tracker.

Thanks!  Take a look at my SuiteSparseBLAS.cmake file.  The logic for finding the Intel MKL BLAS is there.

DrTimothyAldenDavis:
I can try it tonight if I have time.  My logic in SuiteSparseBLAS.cmake is a bit tricky. It tries finding lots of BLAS with a preference for BLAS libraries with 64 bit integer parameters. I've known about the MKL cmake scripts but haven't ever been able to get them to work in SuiteSparse.

DrTimothyAldenDavis:
I'm making progress.  It seems like I can use something like the example you pointed me to.  I need to revise SuiteSparse_config so that it loads OpenMP in the SuiteSparsePolicy.cmake, for all packages, so it can properly detect which OPenMP library will be used.  Once I detect that, I can then set MKL_THREADING and do find_package (MKL ...).

It looks like I can keep the SUITESPARSE_USE_FORTRAN default setting unchnaged, as ON.

DrTimothyAldenDavis:
No luck.  I can add a search for the OpenMP in SuiteSparsePolicy.cmake, and detect that libiomp is in use (by icx).  If I see that, I set MKL_THREADING to 'intel_thread', and then use find_package (MKL ...).  That is supposed to use MKL_THREADING to tell the MKL to link against libiomp.  It doesn't; it links with libgomp instead.

In CHOLMOD:

```
-- The C compiler identification is IntelLLVM 2025.0.4
-- The CXX compiler identification is IntelLLVM 2025.0.4
...
-- Found OpenMP_C: -fiopenmp  
-- Found OpenMP_CXX: -fiopenmp  
-- Found OpenMP: TRUE  found components: C CXX 
-- OpenMP C libraries:   /home/davis/intel/oneapi/compiler/2025.0/lib/libiomp5.so;/lib/x86_64-linux-gnu/libpthread.a 
-- OpenMP C++ libraries: /home/davis/intel/oneapi/compiler/2025.0/lib/libiomp5.so;/lib/x86_64-linux-gnu/libpthread.a 
-- Using Intel OpenMP
-- MKL threading: intel_thread
-- Upgrade to CMake version 3.25.2 or later for native support of Intel compiler(s) icx, icpx. You are running version 3.22.1.
-- MKL_VERSION: 2025.0.1
-- MKL_ROOT: /home/davis/intel/oneapi/2025.0
-- MKL_ARCH: intel64
-- MKL_SYCL_LINK: None, set to ` dynamic` by default
-- MKL_LINK: None, set to ` dynamic` by default
-- MKL_SYCL_INTERFACE_FULL: None, set to ` intel_ilp64` by default
-- MKL_INTERFACE_FULL: None, set to ` intel_ilp64` by default
-- MKL_SYCL_THREADING: intel_thread
-- Using MKL::MKL_SYCL* targets with intel_thread may have potential composability problems on CPU device with other SYCL kernels.
-- MKL_THREADING: intel_thread
-- MKL_MPI: None, set to ` intelmpi` by default
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_scalapack_ilp64.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_cdft_core.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_intel_ilp64.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_intel_thread.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_core.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_blacs_intelmpi_ilp64.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_sycl_blas.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_sycl_lapack.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_sycl_dft.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_sycl_sparse.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_sycl_data_fitting.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_sycl_rng.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_sycl_stats.so
-- Found /home/davis/intel/oneapi/2025.0/lib/libmkl_sycl_vm.so
-- Found /home/davis/intel/oneapi/compiler/latest/lib/libiomp5.so
-- Checking if Fortran is available and compatible with C/C++
-- Looking for a Fortran compiler
-- Looking for a Fortran compiler - /usr/bin/f95
-- The Fortran compiler identification is GNU 11.4.0
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Check for working Fortran compiler: /usr/bin/f95 - skipped
--  
-- Incompatible Fortran/C/C++ compilers detected:
--     Fortran:          /usr/bin/f95
--     Fortran id:       GNU
--     C                 /home/davis/intel/oneapi/2025.0/bin/icx
--     C       id:       IntelLLVM
--     C++               /home/davis/intel/oneapi/2025.0/bin/icpx
--     C++     id:       IntelLLVM
CMake Warning at /home/davis/dev2/SuiteSparse/SuiteSparse_config/cmake_modules/SuiteSparsePolicy.cmake:364 (message):
  Warning: Using Fortran with SuiteSparse requires that it has the same
  compiler ID as the C/C++ compilers.  Use a compatible Fortran compiler, or
  set SUITESPARSE_USE_FORTRAN to OFF.
  CMakeLists.txt:41 (include)

-- Fortran:          enabled
...
-- Found OpenMP_C: -fiopenmp  
-- Found OpenMP: TRUE  found components: C 
-- CHOLMOD has OpenMP: ON
```

Now CHOLMOD looks for the BLAS.  It finds the MKL and uses "find_package (MKL ...)" and before that, MKL_THREADING is set to intel_thread.  So far so good.

```
-- Looking for Intel 32-bit BLAS
-- MKL_THREADING: intel_thread
-- Found Intel MKL
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Looking for Fortran sgemm
-- Looking for Fortran sgemm - found
```

and it fails here.  MKLConfig ignores MKL_THREADING and choose to link against libgomp instead.

```
-- Found BLAS: /home/davis/intel/oneapi/2025.0/lib/libmkl_gf_lp64.so;/home/davis/intel/oneapi/2025.0/lib/libmkl_gnu_thread.so;/home/davis/intel/oneapi/2025.0/lib/libmkl_core.so;/usr/lib/gcc/x86_64-linux-gnu/11/libgomp.so;-lm;-ldl  
-- Looking for Fortran cheev
-- Looking for Fortran cheev - found
-- Found LAPACK: /home/davis/intel/oneapi/2025.0/lib/libmkl_gf_lp64.so;/home/davis/intel/oneapi/2025.0/lib/libmkl_gnu_thread.so;/home/davis/intel/oneapi/2025.0/lib/libmkl_core.so;/usr/lib/gcc/x86_64-linux-gnu/11/libgomp.so;-lm;-ldl;-lm;-ldl  
-- OpenMP C libraries:      /home/davis/intel/oneapi/compiler/2025.0/lib/libiomp5.so;/lib/x86_64-linux-gnu/libpthread.a
-- OpenMP C include:        
-- OpenMP C flags:          -fiopenmp
-- Looking for fmax
-- Looking for fmax - found
-- LAPACK libraries:    /home/davis/intel/oneapi/2025.0/lib/libmkl_gf_lp64.so;/home/davis/intel/oneapi/2025.0/lib/libmkl_gnu_thread.so;/home/davis/intel/oneapi/2025.0/lib/libmkl_core.so;/usr/lib/gcc/x86_64-linux-gnu/11/libgomp.so;-lm;-ldl;-lm;-ldl
-- LAPACK include:      
-- LAPACK linker flags: 
-- BLAS libraries:      /home/davis/intel/oneapi/2025.0/lib/libmkl_gf_lp64.so;/home/davis/intel/oneapi/2025.0/lib/libmkl_gnu_thread.so;/home/davis/intel/oneapi/2025.0/lib/libmkl_core.so;/usr/lib/gcc/x86_64-linux-gnu/11/libgomp.so;-lm;-ldl
-- BLAS include:        /home/davis/intel/oneapi/2025.0/include
-- BLAS linker flags:   
```

So find_package (MKL ...) has the same behavior as find_package (BLAS).  If I turn off Fortran, everything works fine and I get libiomp only.  If I enable Fortran, both MKLConfig.cmake and FindBLAS.cmake insist on using MKL with libgomp instead of libiomp, so I get a CHOLMOD library that links with two OpenMP libraries.

DrTimothyAldenDavis:
The failed attempt is here: [66b027b](https://github.com/DrTimothyAldenDavis/SuiteSparse/pull/989/commits/66b027b4818a7843204171866326c1a6541c5f7f)

DrTimothyAldenDavis:
Assuming all the CI workflows pass, I think I have a workable solution for SuiteSparse 7.11.0, which I can release soon as a stable release.

1. SUITESPARSE_USE_FORTRAN is ON by default, unchanged from the current stable release of SuiteSparse.  This is best because it is the least disruptive.
2. If Fortran is enabled, I issue a fatal error if icx and gfortran are being used, since I know this case will lead to two OpenMP libraries being used, because the Intel MKL BLAS (found by default if icx and oneAPI is installed) will link with the GNU OpenMP but the rest of SuiteSparse will link with the Intel OpenMP.  Otherwise, I issue a warning and tell the user/installer to check if everything is OK (to ensure that only one OpenMP library is in use).  These warnings are issued for some of the CI workflows; all of those are OK because they all use single-threaded BLAS.

I wasn't able to get `find_package (MKL CONFIG)` to resolve this issue; it still insists on linking MKL with libgomp (even if I tell it to use Intel OpenMP by setting MKL_THREADING to 'intel_thread'), probably because it also detects that GNU Fortran is in use.  This is exactly how `find_package (BLAS)` works.

mmuetzel:
Thank you for looking into this in such detail and coming up with a solution that should be fine for most.

Should the explicit enabling of `SUITESPARSE_USE_FORTRAN` in the CI workflows be reverted again?

DrTimothyAldenDavis:
I'd prefer to keep them in the CI so I can see which workflow uses Fortran and which doesn't with a single grep.  So in the future, if want to examine this Fortran / BLAS / OpenMP issue again, I can more easily see where it's used in all the workflows.

PR #1022, opened 2025-12-29
Fix: Allow OpenBLAS detection to proceed on Windows if try_run fails …

matusi143:
Fixed build issue on CMAKE with MSVS 2026.


DrTimothyAldenDavis:
The purpose of this check is to determine which version of OpenBLAS is being used, and how to link it.  If the program fails to run, won't the final application fail to run?  

The 2 checks are supposed to determine if (1) openblas_get_num_threads is available (appears in the OpenBLAS version on March 2015) and (2) if openblas_set_num_threads_local (Apr 2024) is available.  The codes will compile OK either way, because I include the definition of these methods in the two files themselves (check_openblas_Mar2015.c and check_openblas_Apr2024.c).  It's only when I try to run them that the check is actually made.  If the mar2015 one fails to run then OpenBLAS is far too old to use.  If the apr2024 file fails to run then I do not have access to the openblas_set_num_threads_local function, which I need for best performance.

This change defeats the purpose of the tests, so I can't tell which features of OpenBLAS are available when SuiteSparse is compiled.

So I can't use this change as-is.  A better solution would be to figure out why the 2 programs are not running when they should be running.

What version of OpenBLAS do you have?

DrTimothyAldenDavis:
This is the wrong approach.

PR #1027, opened 2026-03-21
Fix gcc16 warning array parameter

gruenich:


gruenich:
@DrTimothyAldenDavis This is an interface change. Do you think it is worth the trouble?

DrTimothyAldenDavis:
Unsure.  I'm worried about backward compatibility with user codes that use the current API.  If I can be sure it doesn't actually break anything, that would be essential.  Otherwise I should play it safe and not change the user API.

mmuetzel:
Afaict, calling conventions are the same for `double *` vs. `double[2]` for function arguments in C for all targets.
The only difference is probably that compilers will be able to analyze a size mismatch while building (e.g., with `-Warray-parameter`).
But there shouldn't be any difference at runtime. (The ABI is unchanged.)

I don't think that this change can break anything downstream.

DrTimothyAldenDavis (commented):
This change seems safe in terms of backward compatibility, and looking over the code I see that I already use "double beta [2]" etc for the related methods (compare cholmod_add with cholmod_l_add for example).  So it would be best to add these changes.

PR #1039, opened 2026-05-01
Instruct dependabot to create pull-requests for the `dev2` branch

mmuetzel:
The processes for the SuiteSparse repository are a bit unusual in that PRs should be opened against the `dev2` branch instead of against the default branch (`dev`).

Instruct dependabot to open the pull-requests for updates of actions in workflow files also against the `dev2` branch. (That might help to avoid merge conflicts when the two branches need to be consolidated inevitably at some point.)


DrTimothyAldenDavis:
Thanks!  This is a great help.

